PR #1, opened 2015-08-25
[WIP] trying to make look work, but the convertion is not the right one

rom1504:
Don't merge


rom1504:
the new broken look try is at #18

PR #33, opened 2015-09-02
Add setBlock to API and /setblock command

demipixel:
Added /setblock (supports relative like /setblock ~ ~-1 ~ 20)

Also added serv.setBlock (prob needs a better name). Sets to world (setBlockType) and sends setBlock to all players (which writes it to the client).


rom1504:
Thanks, that's probably useful.
There are a few things to fix in the resulting code though.

* If setBlock doesn't change the block in the server representation, then changeBlock probably shouldn't either, and there shouldn't be a changeBlock.
* these new functions need to be added to the api.md

I'm not sure yet how that should be done properly, but the way it's currently done make the API hard to understand without looking at the code.

mhsjlw:
Holy regex.

Gjum:
@mhsjlw https://github.com/SpockBotMC/SpockBot/blob/master/spock/utils.py#L128 :smile:

demipixel:
@rom1504 Good point on changeBlock, I'll do this as soon as I can.

demipixel:
@rom1504 Wait hold on. I was rethinking this, and...

changeBlock is the player saying "I've placed it so change it for everybody else and the server"
setBlock is the SERVER saying "Change it for this player"

I feel like we need both...? I don't want to send an extra packet to the player setting a block they've already placed...

rom1504:
Rename player.setBlock to player.sendBlock and add all these new exposed functions to the api (and use player.sendBlock in player.placeBlock) and it's good to merge.

lib/playerPlugins/commands.js
```diff
@@ -11,6 +11,22 @@ function inject(serv, player, options)
...
     }
+    else if(results=command.match(/^setblock/)) { // Like old version which uses ids
+        results = command.match(/^setblock (~|~?-?[0-9]*) (~|~?-?[0-9]*) (~|~?-?[0-9]*) ([0-9]{1,3})/);
+        if(!results) {
+            player.chat("Usage: /setblock <x> <y> <z> <id>");
+        }
+        else {
+            results = results.map(function(num, i) { // parseInt paramaters
+                if (num.indexOf('~') == 0) {
+                    return (Math.floor(player.entity.position[['','x', 'y', 'z'][i]]) + parseInt(num.slice(1) || 0))>>5;
```
rom1504: what is [['','x', 'y', 'z'][i]] ?

lib/playerPlugins/commands.js
```diff
@@ -11,6 +11,22 @@ function inject(serv, player, options)
...
+                }
+                else
+                    return i < 4 ? parseInt(num)>>5 : parseInt(num); // return parseInt>>5 if position, not id
```
rom1504: what is this ? what is it trying to do ?

demipixel:
player.sendBlock + player.changeBlock = serv.setBlock
putting sendBlock in changeBlock just duplicates serv.setBlock. When a player places a block, they don't need a block place event to fire, correct? The server needs to simply send it to others. Otherwise we would just use serv.setBlock.

The whole command support /setblock <x> <y> <z> <id>. It also supports relative positions, so you can do:

/setblock ~ ~-1 ~ 1 to set stone under the player. You can also use a mix of relative and fixed positions (this is all in vanilla minecraft by the way). Since results will be an array, positions 1, 2, and 3 containing "x", "y", and "z", the easiest way to compactly get the player's position (i.e. the ~) is to just do:

player.entity.position[q] where q is ['', 'x', 'y', 'z'][i] where i is what result it's in. This means 1 => x, 2 => y, etc.

demipixel:
Made the changes

demipixel:
Nonononono how do I remove these new commits ;_;

demipixel:
Should I just duplicate my current repo, remove those 3 commits, paste back the repo, commit them all as one, then wait to push until this PR is accepted or what not? Git hates me ;_;

roblabla:
@demipixel here's the procedure (note : make a backup copy of your repo in case shit goes wrong) : 

`git rebase -i fa2e030c55df38faab85958025e0874513bbf226` (this is the current master commit)

Look for your first commit. It should say pick next to it. Now, all the commits under it, instead of "pick"ing them, "squash" them. Save and quit the editor, git should do some work. Once it's done, git push --force and if you did everything right you should have a clean PR. If you did it wrong... that's where the backup copy comes handy ^^'

PR #37, opened 2015-09-05
Update package.json

Creeplays:
Use a precompiled prismarine-chunk and prismarine-world


rom1504:
That's not needed (it already works as-is) and that set the dependency to a precise commit which is not good.

PR #39, opened 2015-09-06
Implemented kicking and banning, as well as /kick and /ban

demipixel:
The worst thing about this is the implementation of `serv.getUUIDFromUsername(username,callback)` but I found nothing else that had it (since the 0x02 login packet contains the UUID so you never need to fetch it externally).

This DOES mean that /ban will take a bit to respond. I don't know any other way to do it since vanilla bans according to UUID, so you need to get the UUID from username if you're doing /ban <username> or /pardon <username>.

I've tested it over and over, so ideally it should work :P


rom1504:
merged there 31188235208574b04960a78ea0b952c30e80aef3

PR #45, opened 2015-09-06
Adding the basics of World Gen

demipixel:
Added a basic "SuperFlat" in world.js. We'll prob make it it's own class, with a folder of default generations. Superflat.js would simply set module.exports instead of "var SuperFlat".

chunkSeed uses https://www.npmjs.com/package/random-seed

That means you can do stuff like chunkSeed.random(), chunkSeed.range(0, 100), and other stuff in that package. The chunk should generate exactly the same every time unless you run Math.random(). <3 seeds!

I presume we'll make a /gens folder soon for some defaults we want to throw around (and maybe event default generation!)

### DIFF
--- a/lib/generations/superflat.js
+++ b/lib/generations/superflat.js
@@ -1,0 +1,29 @@
+var WorldGen = require('../worldgen');
+var Vec3 = require('vec3');
+
+module.exports = WorldGen({
+      'opt': 'default',
+      'bottom_id': 7,
+      'middle_id': 1,
+      'top_id': 2,
+      'middle_thickness': 3,
+      'debug': false
+  }, function(chunk, opt, chunkSeed, chunkX, chunkZ) {
+      var height = opt.middle_thickness + 1;
+      var DEBUG_POINTS = [new Vec3(0, height, 0), new Vec3(15, height, 0), new Vec3(0, height, 15), new Vec3(15, height, 15)];
+      for (var x = 0; x < 16; x++) {
+        for (var z = 0; z < 16; z++) {
+          for (var y = 0; y < opt.middle_thickness + 2; y++) {
+            if (y == 0) chunk.setBlockType(new Vec3(x, y, z), opt.bottom_id);
+            else if (y < opt.middle_thickness + 1) chunk.setBlockType(new Vec3(x, y, z), opt.middle_id);
+            else chunk.setBlockType(new Vec3(x, y, z), opt.top_id); 
+          }
+        }
+      }
+      
+      if (opt.debug) {
+          for (var d in DEBUG_POINTS) {
+              chunk.setBlockType(DEBUG_POINTS[d], 35);
+          }
+      }
+  });
--- a/lib/serverPlugins/world.js
+++ b/lib/serverPlugins/world.js
@@ -1,5 +1,6 @@
 var Chunk = require('prismarine-chunk')(require("../version"));
 var World = require('prismarine-world');
+var WorldGen = require('./worldgen');
 var Vec3 = require('vec3');
 
 module.exports=inject;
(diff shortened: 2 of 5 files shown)


rom1504:
See https://github.com/mhsjlw/flying-squid/issues/47

demipixel:
I did it a bit differently so we could avoid option-name conflicts:

gen(defaultChunkOpt, seed)

gen.generateWorld(genWorldOptions, chunkOptions)

- genWorldOptions contains stuff like sizeX, sizeZ and setLight (set light level 15 everywhere)
- chunkOptions sets the options for each chunk. Currently I'm using `debug: true` which puts wool at the border of every chunk

gen.generate(chunkX, chunkZ, opt)

- opt that are provided to the generation. Normally you won't need this as you can simply set the defaults at the top (defaultChunkOpt).

rom1504:
I can't see why generateWorld should take any option, I think you should be passing the option of size and everything in the opts of the module.

I also can't see the point of separating word option and chunk options, if you really want to have the chunk options in a separate namespace you can always put chunk:{debug:true} in the opts object.

Why do generate (which should really be generateChunk) takes opt as a parameter ? it should use the opt given to the module.

demipixel:
generate() DOES take options from the module, however to allow flexibility, people can chose to put in their own set of options (for example if a plugin wants to force a jungle temple @ 1000, 1000, that's super easy without changing anything).

I don't see the issue with world options vs chunk options? Look at [this](https://github.com/demipixel/flying-squid/blob/world_gen/lib/serverPlugins/world.js#L11). It's also more organized!

defaultChunkOptions => chunkOptions => world gen func (does what it needs with them)

worldChunkOpt => chunkOptions => ...

worldOptions => extra stuff like size, adding lighting, etc

Since chunkOptions is defined by the writer of the generation, we won't know what options will be in it and can't assume "setLight" won't be one of them.

rom1504:
The problem with your current organization is it is too specific, it forces all world generation modules to look like yours.
"for example if a plugin wants to force a jungle temple @ 1000, 1000, that's super easy without changing anything" : for this use case, it seems pretty clear to me the plugin would be assuming a particular world generation, so I guess you can take additional parameter in worldGeneration if you want, but it should still work when providing no parameter.

The issue with putting both these options is that it assumes the module has chunk options (and I might want to put columns options or 10-column options or whatever, I should put that in opts, not create 10 options args)

"Since chunkOptions is defined by the writer of the generation, we won't know what options will be in it and can't assume "setLight" won't be one of them." : yes that's the main point, we want to be able to put any options, the generator module has to document whatever options should be given for that particular world generation module.

demipixel:
I see what you mean... However, it's **extremely** likely that `generateWorld()` will always use some loop through `generate()`. If somebody wants to use multiple options, they should use an array or something.

Think of it this way: You wouldn't write an API where you say "These options only do anything if you call this method with them". What if somebody wants to make a plugin where they shift through multiple seeds to find them best one? If they change the options in the middle, generateWorld will need to have different options; This method makes that easy. Forcing them to come in through the module would require you to create a new generation for every world you make.

This becomes a problem when we have a plugin that, for example, saves an object of really cool, custom generations. Now, if you have 6 players in your minigame and they each get their own mini-world... Well you can't change world options if they were already set by the "cool-generations" now can you! In addition, cool-generations only provides default options for chunks, so if you so choose, you can also generate with custom options.


Point being, we don't know where they'll be used, so we can't assume everybody will have access to creating the generation. This is why everything needs to be mutable, and only defaults are passed in (to make life easier).

rom1504:
Okay so re-reading the conversation here, it seems there is no point arguing more here neither, so I'll just code what I meant in a new branch, hopefully that'll make things clearer.

rom1504:
not touching this until the world implementation is more advanced, there's no point.

rom1504:
Can you rebase this ? (`git fetch` then `git rebase master` then solve the conflicts then `git push --force`)
Thanks

rom1504:
do reopen it though ;)

PR #49, opened 2015-09-07
The Start of Plugins

demipixel:
Well... If all we do is talk about it, we're not gonna get anywhere. Here's a start so at least other people can work on it or test it to see if we actually have any of the problems we were talking about.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,3 @@
 node_modules
+plugins/*
+!plugins/README.md
--- a/doc/api.md
+++ b/doc/api.md
@@ -40,7 +40,7 @@
       - ["disconnected"](#disconnected)
       - ["error" (error)](#error-error-1)
       - ["kicked" (kicker,reason)](#kicked-kickerreason)
-    - [Cancelable Events](#cancelable-events)
+    - [Cancelable Behaviors](#cancelable-behaviors)
       - ["chatMessage"](#chatmessage)
       - ["chat"](#chat)
       - ["command"](#command)
@@ -215,7 +215,7 @@ Fires when there is an error.
 
 `kicker` kick the player with `reason`
 
-### Cancelable Events
+### Cancelable Behaviors
 
 This type of event is emitted by the the player with the option to cancel a default. It is primarily used by external plugins.
 This type of event is emitted twice. For example, if a player digs a block, both digBlock\_cancel and digBlock are emitted.
--- a/doc/api.md
+++ b/doc/api.md
@@ -39,8 +39,17 @@
       - ["spawned"](#spawned)
       - ["disconnected"](#disconnected)
       - ["error" (error)](#error-error-1)
-      - ["chat" (message)](#chat-message)
       - ["kicked" (kicker,reason)](#kicked-kickerreason)
+    - [Cancelable Events](#cancelable-events)
+      - ["chatMessage"](#chatmessage)
+      - ["chat"](#chat)
+      - ["command"](#command)
+      - ["startDig"](#startdig)
+      - ["stopDig"](#stopdig)
+      - ["finishDig"](#finishdig)
+      - ["placeBlock"](#placeblock)
+      - ["attackPlayer"](#attackplayer)
+      - ["animation_arm"](#animation_arm)
     - [Methods](#methods-1)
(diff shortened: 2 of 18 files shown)


rom1504:
In api.MD events should specify their params in parentheses (see new player event for an example)

demipixel:
@roblabla Wanted them all in an object, so every single one would have the paramater (e)

In addition, there are two events for per written one in the API, so really we're looking at:

```
placeBlock_cancel (e, cancel)

placeBlock (e, cancelled)
```

I tried to explain this as best as I could in the description of the section.

rom1504:
Ah OK. I didn't read everything but why do you need to pass 'self' ? What is that ?

demipixel:
For stuff like `player.plugins[self.id].myVariable = 2`. If we find a better way of doing this or if, in the future as we approach beta, we discover it's absolutely useless, we can remove it. It took 3 seconds to add and it'll take less to delete. In addition, idk what "options" is used for yet (or if it is at all), so I'm not too worried about plugins breaking :P

rom1504:
Why not use this instead of passing self ? It's the same object

demipixel:
It's not

https://github.com/demipixel/flying-squid/blob/plugins/lib/serverPlugins/plugins.js#L63

rom1504:
* [ ] remove 'self' (you don't need it), the data independant from a player should be stored in the serv object : you probably want a server plugin for that
* [ ] world guard should definitely be put in a separate repo, not in the plugins dir
* [ ] check that plugins/ dir and plugins.js are needed (pretty sure they are not and people that want to test something quickly should use lib/serverPlugins and lib/playerPlugins)

Overall I can now see how cancelable behaviors can be a way that works to make plugins.

However I'm still convinced there are better ways so it's a good thing that this cancelable behavior pattern don't force you to use them, so I'll be able to make plugins without using them.

demipixel:
- I'm changing self to this by using bind()
- world-guard is in examples
- plugins won't work inside of lib/playerPlugins

rom1504:
* let's see how that works out, but it really seems to me that part of what you are doing in the player plugin should be done in the server plugin part (keep in mind anything you put in a player plugin get executed when each player join)
* I think world-guard could be an useful things in general, why not put it in a separate repo and then links it from the readme as an example ?
* they should. (they do with the current model and they should keep working with that new one)

demipixel:
- I could, however I'm thinking that's gonna make it harder for others to update. It's just something I did in an hour and in no way is it polished (it makes it's own OP system and allows anybody to op themselves!). I'd be happy to put it in it's own repo when we have a final external-plugin design so I don't have to keep making changes myself.
- At the moment, external plugins do not differ between player and server. Most plugins won't need server (such as WG) but in certain cases, when they do need it...

Player plugin:
```js
module.exports = inject; // playerInject
```

Player/Server plugin:
```js
module.exports = {
  player: playerInject,
  server: serverInject
};
```

Server plugin:
```js
module.exports = {
  server: inject // serverInject
}
```

Why? Because nobody wants to turn their 20 line plugin into 2 plugins because it has a serv.on() somewhere.

In addition, "this" cannot be used. I have recently thought of this. That is because:
```js
// somewhere inside the inject method
console.log(this); // The plugin! W00t!
player.on('placeBlock', function(e, cancelled) {
  console.log(this); // The player! Nooooo!
});
```

mhsjlw:
This looks good @rom1504 what do you think? Should we merge it?

mhsjlw:
Merged.

rom1504:
Okay so looking at the code again:
* there shouldn't be a plugins/ dir, plugins.js shouldn't exist: "external" plugins should be loaded the same way "internal" plugins are
* the first point means both kind of plugins should have the same API:
  * self should go away
  * when needed, use a server plugin, not just a player plugin

I don't really like too much the fact that we add tons of complexity to the api and the code with these cancelable methods but I guess if it works, we can keep that until there's a better way.

rom1504:
So after some discussions, it seems the best way here is that I code what I mean in a new branch, it will make things clearer instead of arguing here forever.

rom1504:
@demipixel it might be interesting to port these 2 examples plugins here (noplank and world guard) to the current plugin system

PR #53, opened 2015-09-27
Update README.md

vimes1984:
updating readme to include gulp :D


rom1504:
Actually we should add that kind of thing https://github.com/rom1504/prismarine-chunk/blob/master/package.json#L26 to the devDependencies of flying-squid while we depend on the git url of prismarine-chunk and prismarine-world.

rom1504:
fixed by f5e2b9aaa70196f83eb1b4082aad9871ee08395c

PR #55, opened 2015-09-27
Revert "The Start of Plugins"

rom1504:
Reverts mhsjlw/flying-squid#49


rom1504:
The commits history is pretty messy now, perfect.

PR #57, opened 2015-10-01
The Start of World Generation

demipixel:
Works.

'nuf said.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "prismarine-chunk": "0.2.1",
     "prismarine-entity": "0.1.0",
     "prismarine-world": "0.1.0",
+    "random-seed": "^0.2.0",
     "request-promise": "^0.4.3",
     "requireindex": "~1.0.0",
     "vec3": "0.1.3"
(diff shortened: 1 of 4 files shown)


rom1504:
discussion in #45

demipixel:
So what changes need to be made to this at this point?

rom1504:
Alright so, I think the WorldGen thing is too specific and having a simpler API like just generateChunk is simpler and better.
But your superflat generation works, so could you add it there https://github.com/mhsjlw/flying-squid/tree/master/src/lib/worldGenerations ? (you don't need WorldGen at all for superflat)

rom1504:
Hmm well I guess an options object should be passed to the generation though. (so we can configure the 138 of https://github.com/mhsjlw/flying-squid/blob/master/src/lib/worldGenerations/diamond_square.js#L108 for example)

rom1504:
Alright so I added options. It should be possible to add that superflat.

rom1504:
Okay, done in c4974f8e3388dd8e63843d110d9dd5889717bfc4 , thanks !

PR #60, opened 2015-10-07
fix small bug

JWo1F:


rom1504:
nope, not a bug. You need to npm install for this to work.

PR #62, opened 2015-10-09
fix grass bug

JWo1F:
small fix bug of grass


JWo1F:
Now, I think, it looks very well :)

JWo1F:
Thank you for this project, @rom1504!

PR #67, opened 2015-10-09
Mini worldgen change

demipixel:
Should be good. Added grass, added bedrock/dirt line, matched the waterline (I believe)


rom1504:
can you rebase on master ? (git rebase master then follow the instructions)

rom1504:
Too often it's completely under water

rom1504:
@JWo1F can you look at this ? I'm pretty sure the Math.random doesn't make sense relatively to the diamond square algo.

demipixel:
Math.random stuff is more for population-type things. Minecraft doesn't make worlds exactly the same, it's just the terrain looks similar (it's never exactly the same) and the grass/trees/ores are all in different positions (caves are the same I believe).

If the grass isn't in the same place every time or the bedrock doesn't come up from the bottom exactly the same, nobody is going to notice, and I believe that is how Minecraft does it.

rom1504:
It's not so much about whether minecraft does this that way or not, it's about if the diamond square algorithm does it that way. And it doesn't seem like that.

Why do you need dirtheight ? doesn't DiamondSquare give that already ? ( @JWo1F ?)

demipixel:
No, it was grass and then only dirt underneath. I made it grass, then a few blocks of dirt, then stone (like normal minecraft).

PR #72, opened 2015-10-11
commands

### LINKED ISSUE #42: Improving commands
1. Make server commands work the same as player commands
handleCommand shouldn't assume the command is coming from a player : change player.chat in it by an answer function passed in argument of handleCommand or something like that

2. A better API : there probably should be a `serv.addCommands("setBlock","(~|~?-?[0-9]*) (~|~?-?[0-9]*) (~|~?-?[0-9]*) ([0-9]{1,3}",function(...);` so we can generate the result of the help command easily.

JWo1F:
fix #42

### DIFF
--- a/src/lib/playerPlugins/commands.js
+++ b/src/lib/playerPlugins/commands.js
@@ -1,344 +1,344 @@
 var Vec3 = require('vec3');
 module.exports = inject;
 
-var hash = {};
+function inject(serv, player, options) {
+  var hash = {};
 
-class Command {
-  constructor(params, parent) {
-    this.params = params;
-    this.parent = parent;
-    this.child = [];
+  class Command {
+    constructor(params, parent) {
+      this.params = params;
+      this.parent = parent;
+      this.child = [];
 
-    this.updateHistory();
-  }
-
-  static find(command) {
-    var res;
-    for(var key in hash) {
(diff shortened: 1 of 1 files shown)


rom1504:
Pretty good !
I don't think find and use should be in the Command class though, I think they should be function in inject.
That probably means you will have to put the base.add({ for 'help' in inject too.

JWo1F:
@rom1504 inject call every times when a player connect.

rom1504:
Yeah, that's quite fine. injecting a few things like that will not take any time really

JWo1F:
@rom1504 but what about RAM?

rom1504:
Yeah it might consume a little bit more of ram to do it during player injection. But we could say the same thing about every player._client.on("event",function(...) { ... }); in all the player plugins. 

I think the added simplicity to the code is worth losing a few kB of ram by player.

rom1504:
So, this is supposed to work recursively, right ?
Can you add an example (for example for time) of this http://code.re/8JU working ?

Also I think it would be better to write the usage when people write /help.

And there's a bug there https://github.com/JWo1F/flying-squid/blob/commands/src/lib/playerPlugins/commands.js#L124

rom1504:
(fixed these issues)

Thanks !

PR #73, opened 2015-10-11
Code we won't use

demipixel:
\>.\>

### DIFF
--- a/src/lib/playerPlugins/commands.js
+++ b/src/lib/playerPlugins/commands.js
@@ -3,7 +3,7 @@ var Vec3 = require('vec3');
 module.exports = inject;
 
 function inject(serv, player, options) {
-  function handleCommand(command) {
+  /*function handleCommand(command) {
     var results;
     if (options.commands[command])
       player.chat("" + options.commands[command]);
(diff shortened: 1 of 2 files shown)


rom1504:
The section thing is interesting.

I think we should inspire ourselves from the vanilla help thing.

demipixel:
Lol vanilla doesn't have the section thing, but I figured most plugins would need it.

rom1504:
Hmm what are sections ? I meant the page thing.

demipixel:
Section means it is a command with multiple sub-commands. For example, /time has "set", "add", and "query" (although vanilla doesn't do it this way, it's just an example).

If a plugin, such as world edit, wanted to make a section /we, they would do:

```js
serv.newSection('we', {
  help: 'World Edit Commands'
}
```

Then it's easy to add commands. The section is dedicated by how the command is created. `time` means the section is the default section (i.e. ''), `we time` would mean it was section we> time, `we test time` would mean we > test > time, etc

```js
serv.newSection('we shape', ...);
serv.addCommand('we shape sphere', ...);
serv.addCommand('we shape cube', ...);
serv.addCommand('we fill', ...);
```

As you can see, any section can contain both sub-sections or its own commands. Finally, help commands are automatically created for every single section in the format:

/[section-list] <command> [page-n (optional)]

or

/[section-list] [page-n (optional)]

rom1504:
can probably be used in #42

demipixel:
?????????? Can I delete this and my plugins branch? (old branch from like months ago)

rom1504:
improving the commands is not over and I bet there are still interesting stuff in the plugins branch ;)

rom1504:
Hmm, let's close this. We can always work on it, and the code won't disapear, but it'll be in an other PR anyway.
Let's track work on this in https://github.com/mhsjlw/flying-squid/issues/92

PR #80, opened 2015-10-14
Worlds

demipixel:
I almost cried myself to sleep tonight after I realized I hadn't pulled from master in like 20 commits even though I finished worlds. Luckily you didn't change much :)

- Basics of worlds going. Typing "world" will teleport you to 0, 70, 0 in the nether and generate it if need be.
- A few changes to logging in and setting positions, etc
- serv.worlds, add more worlds by doing serv.worlds.push(new World(...))
- BUG: Time is randomly switching... Is this my fault? Every second it seems to flip back and forth between two "times", meaning it will flash from day to night...? Might be something else, I don't know, because I didn't touch time...
- CODE CRAP: "world" command is stuffed in playerPlugin/world.js. Should I spend the effort of trying to generate a portal both in the overworld (using flint + steel) and in the nether (fully generated by server) or should I simply make a cleaner command until we get around to that?

### DIFF
--- a/doc/api.md
+++ b/doc/api.md
@@ -11,7 +11,9 @@
       - [serv.entityMaxId](#serventitymaxid)
       - [serv.players](#servplayers)
       - [serv.uuidToPlayer](#servuuidtoplayer)
-      - [serv.worlds](#servworlds)
+      - [serv.overworld](#servoverworld)
+      - [serv.netherworld](#servnetherworld)
+      - [serv.endworld](#servendworld)
       - [serv.entities](#serventities)
       - [serv.bannedPlayers](#servbannedplayers)
       - [serv.time](#servtime)
@@ -65,7 +67,7 @@
       - [player.handleCommand(command)](#playerhandlecommandcommand)
       - [player.updateHealth(health)](#playerupdatehealthhealth)
       - [player.sendPosition(position, opt)](#playersendpositionposition-opt)
-      - [player.changeWorld(worldNumber, opt)](#playerchangeworldworldnumber-opt)
+      - [player.changeWorld(world, opt)](#playerchangeworldworld-opt)
     - [Low level properties](#low-level-properties)
       - [player._client](#player_client)
     - [Low level methods](#low-level-methods)
@@ -102,9 +104,17 @@ Array of connected players
 
 Object uuid to players
 
-#### serv.worlds
+#### serv.overworld
 
-Array of all worlds. By default, 0 is Overworld, 1 is Nether, and 2 is the End, however any plugin may push more worlds (`serv.worlds.push(new World(...))`) and the player can be sent there with `player.changeWorld(worldNumber, opt)`.
+Contains the overworld world. This is where the default spoint is and where peope will play survival and such.
+
+#### serv.netherworld
+
+Contains the nether world. This WILL be used when a player travels through a portal if they are in the overworld.
+
+#### serv.endworld
+
+WILL contain the end world. *NOT YET IMPLEMENTED!*
 
 #### serv.entities
 
@@ -230,7 +240,7 @@ The view size of the player, for example 8 for 16x16
 
 #### player.world
 
-Which world the player is in. Get it by using `serv.worlds[player.world]`.
+The world which the player is in.
 
 ### Events
 
@@ -328,9 +338,9 @@ Teleport the player to any `position` in the same world. Use `player.changeWorld
 - yaw: Set the yaw, default is the current yaw of the player
 - pitch: Set the pitch, default is the current pitch of the player
 
-#### player.changeWorld(worldNumber, opt)
+#### player.changeWorld(world, opt)
 
-Change which world the player is in (Overworld, Nether, End, or any other world added to `serv.worlds`). Options:
+The world object which the player is in (use serv.overworld, serv.netherworld, serv.endworld, or a custom world). Options:
 
 - gamemode: Gamemode of the world (Default is player gamemode)
 - difficulty: Difficulty of world. Default is 0 (easiest)
--- a/doc/api.md
+++ b/doc/api.md
@@ -31,6 +31,7 @@
       - [serv.broadcast(message[,color])](#servbroadcastmessagecolor)
       - [serv.setBlock(position,blockType)](#servsetblockpositionblocktype)
       - [serv.getPlayer(username)](#servgetplayerusername)
+      - [serv.getNearby(loc)](#servgetnearbyloc)
       - [server.banUsername(username,reason,callback)](#serverbanusernameusernamereasoncallback)
       - [server.ban(uuid,reason)](#serverbanuuidreason)
       - [server.pardonUsername(username,callback)](#serverpardonusernameusernamecallback)
@@ -38,6 +39,10 @@
       - [server.getUUIDFromUsername(username,callback)](#servergetuuidfromusernameusernamecallback)
       - [server.setTime(time)](#serversettimetime)
       - [server.setTickInterval(ticksPerSecond)](#serversettickintervaltickspersecond)
+    - [Low level methods](#low-level-methods)
+      - [server._writeAll(packetName, packetFields)](#server_writeallpacketname-packetfields)
+      - [server._writeArray(packetName, packetFields, playerArray)](#server_writearraypacketname-packetfields-playerarray)
+      - [server._writeNearby(packetName, packetFields, loc)](#server_writenearbypacketname-packetfields-loc)
   - [Player](#player)
     - [Properties](#properties-1)
       - [player.entity](#playerentity)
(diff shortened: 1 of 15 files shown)


rom1504:
Pretty good.
A few points I think might improve the code:
1. make serv.worlds an object instead of an array so we can so serv.worlds["overworld"] instead of serv.world[0]
2. instead of doing serv.worlds[player.world] everywhere, what about putting a reference of the world directly in player.world ? (it won't copy the world, js works with references for objects)
3. if player.world is the actual world, make a player.worldName to store the world the player is currently in. (maybe the name of the world should be directly in prismarine-world though, but we can change that later)

Well I'm not entirely sure about 1 and 3. I think 2 would improve things though.

demipixel:
Okay, when I do a rebase, it works fine, but then it says "1 behind 19 ahead" or whatever. After the rebase, I had to pull and then merge with origin/worlds or whatever it was, otherwise it wouldn't let me push (obviously).

My biggest issue with your first point is that it's similar to the plugin problem: I'm worried about assuming no two plugins are going to try and take the same name (even though I know serv.worlds.overworld would be cleaner to work with). Sure, we could have a serv.createWorld to do this for us, but that means if two plugins have never met before and use the same world name, ONE of them will crash.

For your second option: I was thinking about that. As long as it's easy to use or compare, I think it would be fine (compare as in playerA.world == playerB.world). I don't know if we'd actually need three if we did this one.

rom1504:
"it wouldn't let me push " : do git push --force (after a rebase)

Yeah I guess we could compare that way.

rom1504:
Ok I just tested it.
That nether is fun.

Introduced bug: can't see the other players. Fix it ;)

As a general rule, when you change stuff in flying-squid, think about testing it with 2 clients, to make sure multiplayer stuff still work ;)

src/lib/playerPlugins/world.js
```diff
@@ -0,0 +1,124 @@
...
+  function spiral(arr)
+  {
+    var t=[];
+    spiralloop(arr,function(x,z){
+      t.push([x,z]);
+    });
+    return t;
+  }
+
+	function sendChunksAroundPlayer(view, world)
```
rom1504: please fix the indentation there ;)

src/lib/playerPlugins/updatePositions.js
```diff
@@ -82,4 +82,23 @@ function inject(serv,player)
     player.entity.onGround = onGround;
     player.emit("positionChanged");
   }
+
+  function sendPosition(pos, opt) {
```
rom1504: what is this opt for ?

src/lib/playerPlugins/login.js
```diff
@@ -243,8 +145,6 @@ function inject(serv,player)
     updateTime();
     setGameMode(player.gameMode);
     fillTabList();
-    spawnOthers();
-    spawn();
```
rom1504: pretty sure you can't do that and the order of stuff need to be respected.

src/lib/worldGenerations/nether.js
```diff
@@ -0,0 +1,31 @@
+var Chunk = require('prismarine-chunk')(require("../version"));
+var Vec3 = require('vec3');
+
+function generation({level=50}={}) {
+	function generateChunk(chunkX, chunkZ) {
```
rom1504: same remark here: fix the indentation

src/lib/playerPlugins/login.js
```diff
@@ -40,78 +40,30 @@ function inject(serv,player)
...
 
     player.on("positionChanged",function(){
       if(!player.sendingChunks && player.entity.position.distanceTo(player.lastPositionChunkUpdated)>16*32)
       {
         player.sendingChunks=true;
-        sendChunksAroundPlayer(player.view).then(() => player.sendingChunks=false);
+        player.sendChunksAroundPlayer(player.view).then(() => player.sendingChunks=false);
       }
     });
+    player.on("teleport", function() {
```
rom1504: No, that's not the way to do it. When a player teleport, its position should be changed (see updatePositions.js), which will then emit positionChanged, and send him the right chunks.

src/lib/playerPlugins/world.js
```diff
@@ -0,0 +1,124 @@
+var vec3=require("vec3");
+var spiralloop = require('spiralloop');
+
+module.exports = inject;
+
+function inject(serv, player) {
+	function spawn() {
```
rom1504: that's wrong.

demipixel:
> 1) please fix the indentation there ;)

k

> 2) Ah and in the rest of the file, we do 2 spaces indentation, configure your editor to do that.

It is configured... o_O I'm using sublime... Sometimes it changes by file (e.g. if it recognizes a 4-space file)

> 3) what is this opt for ?

It's in the api.md. opt.yaw and opt.pitch

> 4) pretty sure you can't do that and the order of stuff need to be respected.

So you want player.spawn() and the bottom again?

> 5) same remark here: fix the indentation

Ahhhhh, I DID have it on two-space display, but I didn't always have it on "tabs=>spaces". *sigh* Sublime...

> 6) No, that's not the way to do it. When a player teleport, its position should be changed (see updatePositions.js), which will then emit positionChanged, and send him the right chunks.

positionChanged should really be named "moved". The difference I'm making is that "moved" is a relativePositionChange and "teleport" is sent by the server via sendPosition. The reason for doing this is because if we have something like this:

```js
    player.on("positionChanged",function(){
      if(!player.sendingChunks && player.entity.position.distanceTo(player.lastPositionChunkUpdated)>16*32)
      {
        player.sendingChunks=true;
        player.sendChunksAroundPlayer(player.view).then(() => player.sendingChunks=false);
      }
    });
```
If they're at 0,0 in the overworld and then teleport to 0,0 in the nether, it all screws up. (Sure, we could set player.lastPositionChunkUpdated to something else, but we have to assume somebody else might have this problem as well). Differentiating between "I've moved some" VS "I moved to another world or moved 100000 blocks" should be noted.

> 7) that's wrong.

You mean indentation again?

rom1504:
3)
Right, why is there even any parameters at all in sendPosition ? I think it would be better to just set the player.entity.position and player.entity.yaw and player.entity.pitch and then just call sendPosition.

4) I want what was at the bottom to be at the bottom, that new player.spawn is wrong (see 7) (see the difference in https://github.com/mhsjlw/flying-squid/pull/80/files )

6) No, it really should be the same thing for any position in a given world. For a change of world, we might have to adapt that condition yeah, pretty sure it's not needed to make a new event for that.

7) no, look at the previous code, this doesn't do the same thing (named_entity_spawn and all are missing there)

src/lib/serverPlugins/world.js
```diff
@@ -7,14 +7,32 @@ var generations={
...
+  'nether':require("../worldGenerations/nether")
 };
 
 module.exports = inject;
 
-function inject(serv,{regionFolder,generation={"name":"diamond_square","options":{"worldHeight":80}}}={}) {
+function inject(serv,{generation={"name":"diamond_square","options":{"worldHeight":80}}}={}) {
   generation.options.seed=generation.options.seed || Math.random()*Math.pow(2, 32);
   serv.emit("seed",generation.options.seed);
-  serv.world = new World(generations[generation.name](generation.options),regionFolder);
```
rom1504: add back the regionFolder please ;)

doc/api.md
```diff
@@ -98,9 +102,9 @@ Array of connected players
 
 Object uuid to players
 
-#### serv.world
+#### serv.worlds
```
rom1504: didn't you remove .worlds ?

src/lib/worldGenerations/nether.js
```diff
@@ -0,0 +1,31 @@
+var Chunk = require('prismarine-chunk')(require("../version"));
+var Vec3 = require('vec3');
+
+function generation({level=50}={}) {
+  function generateChunk(chunkX, chunkZ) {
+    var chunk=new Chunk();
+    for (var x = 0; x < 16; x++) {
+      for (var z = 0; z < 16; z++) {
+        var bedrockheighttop = 1 + Math.round(Math.random()*3)
```
rom1504: https://github.com/mhsjlw/flying-squid/issues/82

src/lib/serverPlugins/world.js
```diff
@@ -15,6 +16,23 @@ module.exports = inject;
...
-  serv._worldSync=new WorldSync(serv.world);
-}
+  serv.overworld = new World(generations[generation.name](generation.options), regionFolder);
+  serv.netherworld = new World(generations["nether"]({}));
+  //serv.endworld = new World(generations["end"]({}));
+
+  //serv._worldSync=new WorldSync(serv.worlds[0]);
```
rom1504: `serv._worldSync=new WorldSync(serv.overworld);` should do it

src/lib/playerPlugins/updatePositions.js
```diff
@@ -82,4 +82,24 @@ function inject(serv,player)
...
     player.emit("positionChanged");
   }
+
+  function setPosition(pos, opt) {
+    opt = opt || {};
+    if (pos) player.entity.position = toFixedPosition(pos);
+    if (typeof opt.yaw != 'undefined') player.entity.yaw=opt.yaw;
+    if (typeof opt.pitch != 'undefined') player.entity.pitch=opt.pitch;
+    
+    player._client.write('position', {
```
rom1504: just keep in mind that this won't move the player after he spawned (got to use the teleport packet for that), maybe the name of the function should reflect that (maybe setInitialPosition)

rom1504:
Working and the code is ok, just need to do the "spawn player if close enough" since we don't send to everybody the position of everybody now.

rom1504:
Finally done :D 
I added commits to fix the remaining issues.

PR #89, opened 2015-11-06
Sound and Particles!

demipixel:
'nough said


rom1504:
hmm maybe add these 3 new serv. function in the api.md ?

PR #94, opened 2015-11-08
Entities and Physics

demipixel:
In progress

### DIFF
--- a/src/lib/playerPlugins/digging.js
+++ b/src/lib/playerPlugins/digging.js
@@ -77,13 +77,10 @@ function inject(serv,player)
     if(expectedDiggingTime-diggingTime<100) {
       player.changeBlock(location,0,0);
       // Drop block
-      var vec = Vec3(location.x, location.y, location.z);
-      var id = await player.world.getBlockType(vec);
-      var damage = await player.world.getBlockData(vec);
-      serv.spawnObject(2, player.world, vec.clone().add(Vec3(0.5, 0.5, 0.5)), {
+      serv.spawnObject(2, player.world, location.offset(0.5, 0.5, 0.5), {
         velocity: Vec3(Math.random()*4 - 2, Math.random()*2 + 2, Math.random()*4 - 2),
-        itemId: id,
-        itemDamage: damage
+        itemId: currentlyDugBlock.type,
+        itemDamage: currentlyDugBlock.metadata
       });
     }
     else


rom1504:
Things to change:
* [ ] remove all these remaining console.log debug code
* [ ] make block drop not fly

src/lib/playerPlugins/updatePositions.js
```diff
@@ -47,7 +47,7 @@ function inject(serv,player)
 
   function sendRelativePositionChange(newPosition, onGround) {
     if (player.entity.position.distanceTo(new vec3(0, 0, 0)) != 0) {
-      var diff = newPosition.minus(player.entity.position);
+      var diff = newPosition.scaled(1/32).scaled(32).minus(player.entity.position.scaled(1/32).scaled(32));
```
rom1504: that was a test, remove it ;)

rom1504:
I tested the last commit on my computer, I'm not getting that tp bug we were getting. Remaining issues I saw:
* [ ] block drop respawn as stone (I guess you forgot to resend the entity metadata packet)
* [ ] updateAndSpawn needs to be called for every entity moving not just players

src/lib/playerPlugins/communication.js
```diff
@@ -8,15 +8,17 @@ function inject(serv,player)
       .forEach((otherPlayer) => otherPlayer._client.write(packetName, packetFields));
 
   player._writeOthersNearby = (packetName, packetFields) =>
-    serv._writeArray(packetName, packetFields, player.nearbyPlayers);
+    serv._writeArray(packetName, packetFields, player.nearbyPlayers());
```
rom1504: no, there is not (), that's indeed player.nearbyPlayers

rom1504:
Alright it's mostly done. Mostly docs remaining now.

src/lib/playerPlugins/logout.js
```diff
@@ -17,7 +21,7 @@ function inject(serv,player)
           UUID: player._client.uuid
         }]
       });
-      player.nearbyPlayers.forEach(otherPlayer => otherPlayer.despawnPlayers([player]));
+      player.nearbyPlayers().forEach(otherPlayer => otherPlayer.despawnPlayers([player]));
```
rom1504: there is no (), that's the point of player.nearbyPlayers

src/lib/playerPlugins/communication.js
```diff
@@ -8,15 +8,17 @@ function inject(serv,player)
...
-      radius: player.playerViewDistance*32
-    })
-    .filter((p) => p != player);
+  player.getNearbyPlayers = (radius=player.entity.viewDistance*32) => serv.getNearby({
+    world: player.world,
+    position: player.position,
+    radius: radius
+  });
+
+  player.nearbyPlayers = (radius=player.entity.viewDistance*32) => player.entity.nearbyEntities
```
rom1504: the radius cannot be used here, so remove it ;)

second point : it's really confusing to have nearbyPlayers() and getNearbyPlayers()

it would be better to make it getNearbyPlayers() and computeNearbyPlayers()

src/lib/serverPlugins/entities.js
```diff
@@ -0,0 +1,280 @@
...
+      playersToAdd.forEach(p => p.spawnEntity(entity));
+      playersToAdd.forEach(p => p.entity.nearbyEntities=p.entity.getNearby());
+
+      entity.nearbyEntities=updatedEntities;
+    };
+
+    return entity;
+  }
+
+  serv.spawnObject = (type, world, position, {pitch=0,yaw=0,velocity=vec3(0,0,0),data=1,itemId,itemDamage=0}={}) => {
```
rom1504: better call itemId,itemDamage as type and metadata as we do in the rest of flying-squid

rom1504:
About nearbyPlayers : if we need it, we should keep it as an array, not a function and simply update it in updateAndSpawn (by filtering on nearbyEntities)

rom1504:
Got to fix changing world @demipixel , it's broken.

PR #108, opened 2015-11-20
Just like how caterpillars will turn into butterflies...

demipixel:
Players will turn into Entities.


109C:
In one ~~sweep~~ pull request the player is not longer unique in it's own class, being to reduced to a mere entity... Beautiful.

src/lib/plugins/blocks.js
```diff
@@ -1,5 +1,3 @@
-var Vec3 = require("vec3").Vec3;
```
rom1504: That line is needed (see the new Vec3 down there)

src/lib/plugins/entities.js
```diff
@@ -76,6 +73,16 @@ module.exports.server=function(serv,options) {
         if (entity.deathTime && Date.now() - entity.bornTime >= entity.deathTime) {
           entity.destroy();
           return;
+        } else if (entity.pickupTime && Date.now() - entity.bornTime >= entity.pickupTime) {
+          var players = serv.getNearby({
+            world: entity.world,
+            position: entity.position,
+            radius: 1.5*32 // Seems good for now
+          });
+          if (players.length) {
```
rom1504: Why is that check needed ? UpdateAndSpawn already handle counting players and all.

src/lib/plugins/pvp.js
```diff
@@ -15,17 +15,17 @@ module.exports.player=function(player,serv)
     if (!serv.entities[entityId]) return; // ?????
     var attackedPlayer = serv.entities[entityId].player;
```
rom1504: entity.player doesn't exist anymore

src/lib/plugins/entities.js
rom1504: Ah nevermind, I didn't understand what you were doing. You implemented collect, fun.

src/lib/behavior.js
```diff
@@ -0,0 +1,16 @@
+module.exports = (obj) => {
+  return (eventName, data, func, opt) => {
+    var hiddenCancelled = false;
+    var cancelled = false;
+    var cancel = (hidden) => { // Hidden shouldn't be used often but it's not hard to implement so meh
```
rom1504: what is hidden cancelled ?

src/lib/plugins/blocks.js
demipixel: Oops

src/lib/plugins/pvp.js
demipixel: Damnit. I still actually need to test the knockback stuff, haha.

src/lib/behavior.js
demipixel: It cancels the behavior without setting "cancelled" to true. This means you can trick other plugins into doing things. Really easy to implement for now, so if we don't find any use cases we can remove it.

rom1504: ok

PR #110, opened 2015-11-21
Come with me, and you'll be...

demipixel:
... in a worllllld of [pure plugins!](https://youtu.be/r2pt2-F2j2g?t=27s)

To test:

1. Clone this branch or whatever
2. Go to plugins folder, create redditbuilds
3. Add an index.js with [this](http://hastebin.com/ofokacuwer.js) and `npm init` the folder
4. Go to /config, duplicate default_settings.json into settings.json
5. Add a plugin `"redditbuilds": {}` inside "plugins"
6. ???
6. Profit. Probably. Hopefully.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ dist/
 modpePlugins/
 config/settings.json
 logs/
+src/plugins/*
+!src/plugins/README.md
--- a/app.js
+++ b/app.js
@@ -21,7 +21,8 @@ var options = {
   generation:settings.generation,
   'modpe': settings.modpe,
   kickTimeout: settings.kickTimeout ? settings.kickTimeout : 10*1000,
-  regionFolder: settings.regionFolder
+  regionFolder: settings.regionFolder,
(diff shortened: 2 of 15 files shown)


rom1504:
I don't understand why you need to create a folder inside plugins/

demipixel:
Because a plugin is a module. That way, plugins can install any other modules they want. Lastly, it makes it extremely easy to npm publish with no issues.

rom1504:
Any file that does module.exports = {} is a module. But okay fine.

demipixel:
@rom1504 I think it would work without the folder, however if I were teaching someone about this, I would teach them to create a folder since it's best practice and since they'll hopefully be publishing it.

rom1504:
Yeah yeah sure, I agree.

109C:
@demipixel Also, it makes plugins/ cleaner if people want to split their plugin into multiple files.

rom1504:
Alright, it works fine, nice.

PR #111, opened 2015-11-22
Implement inventories for players

109C:
So far inventories are kept track of server-side, and slot setting and switching works. Also, players have an inventory full of iron shovels to begin with. How fun.


109C:
Closing because redundant, mineflayer already has a window class.

PR #112, opened 2015-11-22
Use mineflayer's window classes in player's inventory

109C:
~~Not yet finished all the inventory interaction stuff.~~
All finished.

- [x] Add items on set_creative_slot
- [x] Remove items when items are removed in creative
- [x] Pick up items from the ground if there is room
- [x] Allow stacks of items
- [x] Stop the drops from being destroyed when there isn't room. (Not sure where that happens)
- [x] Stack items when they are picked up


src/lib/plugins/entities.js
```diff
@@ -257,13 +270,19 @@ module.exports.entity=function(entity,serv){
   };
 
   entity.collect = (collectEntity) => {
-    if (entity.type != 'player') serv.emit('error', 'Non-player entity (ttype ' + entity.type + ') cannot collect another entity');
-    else {
+    if (entity.type != 'player'){
+      serv.emit('error', 'Non-player entity (ttype ' + entity.type + ') cannot collect another entity')
+    }
```
demipixel: Return here? I think emitting the error logs it but doesn't crash the program

109C: @demipixel I just moved the error thing to another line?

demipixel: Well the "else" meant it wouldn't run the collect code if it threw an error. So that means you have to add a return inside the "if" if you're not using "else".

package.json
```diff
@@ -34,6 +34,8 @@
     "prismarine-entity": "0.1.0",
     "prismarine-world": "0.3.3",
     "prismarine-world-sync": "0.1.0",
+    "prismarine-item": "git://github.com/PrismarineJS/prismarine-item.git",
+    "prismarine-windows": "git://github.com/PrismarineJS/prismarine-windows.git",
```
rom1504: put the release number here instead ( "0.0.0" ) : for both prismarine-item and prismarine-windows

src/lib/plugins/entities.js
```diff
@@ -1,5 +1,6 @@
 var Entity=require("../entity");
 var Vec3 = require("vec3").Vec3;
+var ItemStack = require("prismarine-item")("1.8")
```
rom1504: use require("../version") instead of "1.8" (so we don't have to change the version number everywhere when we move to 1.9)

src/lib/plugins/entities.js
```diff
@@ -170,6 +171,18 @@ module.exports.entity=function(entity,serv){
     if(entity.position.distanceTo(entity.lastPositionPlayersUpdated)>2*32)
       entity.updateAndSpawn();
   });
+  
+  entity.on("inventoryChange", function(){
```
rom1504: this should be in a player plugins
and use the "windowUpdate" event instead of creating a new "inventoryChange" event (see https://github.com/PrismarineJS/prismarine-windows#window-windowupdate-slot-olditem-newitem )

src/lib/plugins/inventory.js
```diff
@@ -1,26 +1,32 @@
...
+var ItemStack = require("prismarine-item")("1.8")
+
 module.exports.player=function(player)
 {
-  player.heldItemSlot=0;
-  player.heldItem=0;
-  player.inventory=new Array(44);
+  player.heldItemSlot = 0
+  player.heldItem = new ItemStack(256, 1)
+  player.inventory = new Windows.InventoryWindow(0, "???", 44)
```
rom1504: the title is "Inventory"

src/lib/plugins/entities.js
```diff
@@ -257,13 +270,20 @@ module.exports.entity=function(entity,serv){
...
 
   entity.collect = (collectEntity) => {
-    if (entity.type != 'player') serv.emit('error', 'Non-player entity (ttype ' + entity.type + ') cannot collect another entity');
-    else {
+    if (entity.type != 'player'){
+      serv.emit('error', 'Non-player entity (ttype ' + entity.type + ') cannot collect another entity')
+      return;
+    }
+    
+    var EmptySlot = entity.inventory.firstEmptyInventorySlot()
```
rom1504: better call it emptySlot. Identifiers starting by a capital letter are class name.

package.json
109C: The reason I put the repository is because they weren't published when I wrote this.

demipixel:
> Stop the drops from being destroyed when there isn't room. (Not sure where that happens)

It happens in entities.js https://github.com/mhsjlw/flying-squid/blob/master/src/lib/plugins/entities.js#L74

After player = serv.getNearby(...) do .filter() and then filter out players with an inventory that cannot pick up that type of item (they could have a full inventory but still be allowed to pick up an item into a slot of the same type).

src/lib/plugins/inventory.js
```diff
@@ -1,26 +1,35 @@
...
   player._client.on("set_creative_slot", ({slot,item} ={}) => {
-    player.inventory[slot]=item;
+    if(item.blockId == -1){
+      player.inventory.updateSlot(slot, undefined)
+      player.emit("windowUpdate", slot, player.inventory.slots[slot], undefined)
+      return;
+    }
+    
+    var NewItem = new ItemStack(item.blockId, item.itemCount, item.metadata)
+    player.emit("windowUpdate", slot, player.inventory.slots[slot], NewItem)
```
rom1504: nah, the whole point of using that event is that it is already emitted by prismarine-windows (see https://github.com/PrismarineJS/prismarine-windows/blob/master/lib/windows.js#L203)

src/lib/plugins/entities.js
```diff
@@ -257,13 +259,22 @@ module.exports.entity=function(entity,serv){
...
+    var emptySlot = entity.inventory.firstEmptyInventorySlot()
+    if(emptySlot != null){
       collectEntity._writeOthersNearby('collect', {
         collectedEntityId: collectEntity.id,
         collectorEntityId: entity.id
       });
       entity.playSoundAtSelf('random.pop');
+      
+      var NewItem =  new ItemStack(collectEntity.itemId, 1, collectEntity.damage)
+      entity.emit("windowUpdate", emptySlot, undefined, NewItem)
```
rom1504: no need to emit it

src/lib/plugins/inventory.js
```diff
@@ -53,4 +60,17 @@ module.exports.player=function(player)
             });
 
   });
+  
+  player.inventory.on("windowUpdate", function(){
+    var Items = entity.inventory.slots
+    
+    for(var ItemIndex in Items){
+      var Item = Items[ItemIndex]
+      entity._client.write("set_slot", {
```
rom1504: player

src/lib/plugins/inventory.js
```diff
@@ -53,4 +60,17 @@ module.exports.player=function(player)
             });
 
   });
+  
+  player.inventory.on("windowUpdate", function(){
+    var Items = entity.inventory.slots
```
rom1504: player not entity

src/lib/plugins/entities.js
```diff
@@ -257,13 +258,37 @@ module.exports.entity=function(entity,serv){
...
   entity.collect = (collectEntity) => {
-    if (entity.type != 'player') serv.emit('error', 'Non-player entity (ttype ' + entity.type + ') cannot collect another entity');
-    else {
+    if (entity.type != 'player'){
+      serv.emit('error', 'Non-player entity (ttype ' + entity.type + ') cannot collect another entity')
+      return;
+    }
+    
+    // Add it to a stack already in the player's inventory if possible
+    for(Item in entity.inventory.items()){
```
rom1504: it's for(var item

src/lib/plugins/entities.js
```diff
@@ -257,13 +258,37 @@ module.exports.entity=function(entity,serv){
...
+    // If we couldn't add it to a already existing stack, put it in a new stack if the inventory has room
+    var emptySlot = entity.inventory.firstEmptyInventorySlot()
+    if(emptySlot != null){
       collectEntity._writeOthersNearby('collect', {
         collectedEntityId: collectEntity.id,
         collectorEntityId: entity.id
       });
       entity.playSoundAtSelf('random.pop');
+      
+      var NewItem =  new ItemStack(collectEntity.itemId, 1, collectEntity.damage)
```
rom1504: newItem. capital letter is for classes

src/lib/plugins/inventory.js
```diff
@@ -53,4 +60,17 @@ module.exports.player=function(player)
             });
 
   });
+  
+  player.inventory.on("windowUpdate", function(){
+    var Items = player.inventory.slots
```
rom1504: items

src/lib/plugins/inventory.js
```diff
@@ -53,4 +60,17 @@ module.exports.player=function(player)
...
+    
+    for(var ItemIndex in Items){
```
rom1504: itemIndex and items

109C:
I'm so used to using the lazy
```
for(thing in things)
```
syntax

rom1504:
if you're really lazy, use things.forEach(thing => { });

109C:
Nah, I prefer the python way, I can read it better.

rom1504:
Okay, but the var is needed, if you forget it, the prog crashes in node 0.10
also start with a lower case letter for anything but classes.

src/lib/plugins/inventory.js
```diff
@@ -53,4 +60,17 @@ module.exports.player=function(player)
...
+  
+  player.inventory.on("windowUpdate", function(){
+    var items = player.inventory.slots
+    
+    for(var itemIndex in items){
+      var item = items[itemIndex]
+      player._client.write("set_slot", {
+        windowId: 0,
+        slot: itemIndex,
+        item: itemStack.toNotch(Item)
```
rom1504: nah itemStack is indeed ItemStack (might want to call it Item though)

rom1504:
Once you're done fixing stuff. Could you git rebase --i 9396e03f1ba3a485e42b6ed538c77da00621186c^ ? (see  https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Changing-Multiple-Commit-Messages ) (and then git push --force)

To clean up the commits (just squash everything in one commit)

109C:
Should be rebased now.

rom1504:
rebased and merged in 28343dc3d365034cd16f8792bf6330a2d5b0364b

PR #122, opened 2015-11-24
No more crashes when using uniplemented windows.js features

109C:


rom1504:
Okay, proper fix got to be done there https://github.com/PrismarineJS/prismarine-windows/issues/1 though

PR #123, opened 2015-11-24
Disable shift clicking

109C:


rom1504:
I bet you can still make it crash, but well, let's merge.

Issue #118, opened 2015-11-23
Chat colors and "links"


demipixel:
Okay, so you can do a lot of things in chat. For basic examples it would be nice to use strings:

`"Hello§1Red"` or better yet `"Hello" + CHAT.RED + "Red"`

Same with others colors, bold, italics, underline, cross-out, random, etc. This string would be converted to JSON for you.

Then there comes more complicated things... Like something happens when you click on text or what text appears when you hover over text. This would most likely be done with JSON manually or something.

rom1504:
How do you do this in vanilla ?

109C:
@demipixel `&` is the informal standard to use colors, while you cannot type `§` into the chat.

@rom1504 It's sent in this format:
```
{
  text: "thanks for the good bones",
  color: "white",
  bold: true,
  extra: [
    {
      text: "mr skeletal"
    }
  ]
}
```

rom1504:
I'm not asking how it is sent, I'm asking how to you send that kind of stuff from the chat in the game. (so it is displayed with the colors and all)

demipixel:
@109C I don't want it usable in chat (plugins can allow that if they want)

@rom1504 /tellraw command (look at mc wiki)

rom1504:
"I don't want it usable in chat" : where do you want to use it then ?

demipixel:
In places like broadcast or broadcastJSON or something. For example, in my redditbuilds.js, it's very hard to read the text. It would be nice to color it to organize it a bit. (Plus we'll need it if we ever implement tellraw). Colors/Bold/Etc also work on signs, btw. (Users still can't do this but plugins do).

rom1504:
hmm, is it really a good idea to use that kind of text format instead of json ? Can't we directly build the json or have helper functions to help us build that json ?

109C:
It's easier to read and build `hi&nthere` than `{text:"hi", extra:{text:"there", underline:true}}`, both for plugins and players. (Which are the same thing, actually)

demipixel:
Yes that's what I meant. Allow something like broadcast("This is in white. &rThis is in red")

and

broadcast("This is in white. \&rThis has funny symbols before it");

As well as broadcastJSON for the format @109C mentioned. You need that for more complicated things like hovering text or clicking to do something.

rom1504:
Done, see #167 for remaining fixes.

PR #130, opened 2015-11-25
Added serv.broadcastColor, which can take '&' color codes

109C:
Does what was suggested in #118.


109C:
Closing PR, I think demipixel's PR is better.

rom1504:
Actually I think your idea of putting this in a separate package is a good idea.
But demipixel PR has more feature, so I think his features should be put in your package and then we can use your package here, like this PR.

PR #133, opened 2015-11-26
proof of concept to find listeners of a plugin

rom1504:


109C:
@rom1504 What if the plugins removes the `newListener` listener? Wouldn't this not catch the subsequent event listener adds?

rom1504:
@109C sure, but plugins can do lot of more malicious things than that (for example player._client=null;)

109C:
@rom1504 It's not that I think the plugin will be malicious, I think that it's too easy to shoot yourself in the foot with it.

rom1504:
I can't see how a plugin would remove `newListener` listener by accident.

rom1504:
anyway, as said in https://github.com/mhsjlw/flying-squid/issues/83#issuecomment-160008204 we probably will want to add listener after loading, so being able to get them at loading is not so useful

PR #140, opened 2015-12-01
Fixed friction bug

109C:


rom1504:
indeed, merging but friction/drag should be implemented properly, see https://github.com/mhsjlw/flying-squid/issues/136#issuecomment-160332220

rom1504:
@109C you should create new branches (without git checkout -b new_branch) so you can make multiple PRs

PR #143, opened 2015-12-06
Add teleport ( aka. tp ) command.

azastrael:


rom1504:
we have a player.sendPosition() function https://github.com/mhsjlw/flying-squid/blob/b5d22e57f93993325d8c96a28ca1ee6ab8a70173/src/lib/plugins/updatePositions.js#L90

So it would be better to 1. change the player position (`player.position = new Vec3(x,y,z)` and then call it `player.sendPosition()`)

rom1504:
Yes, it's planned to abstract it away https://github.com/mhsjlw/flying-squid/issues/40

And we should probably document the reason why 2 position system are used, I'm tired of explaining it again and again and again.

rom1504:
I merged this btw, thanks @azastrael !

PR #144, opened 2015-12-06
Over Powered

demipixel:
added OP to flying-squid (and other shizzles)


src/lib/plugins/tp.js
```diff
@@ -19,14 +26,22 @@ module.exports.player = (player, serv) => {
...
+        
+        player.teleport(new Vec3(x, y, z));
       } else if(args.length === 4) {
         let player_from;
 
         if(!(player_from = serv.getPlayer(args[0])))
           return false;
 
-        player_from.teleport(new Vec3(args[1]*32, args[2]*32, args[3]*32));
+        let x = getPos(args[0], 'x', player_from);
```
rom1504: that's wrong.

rom1504:
I fixed that little mistake and added a regression test to make sure it doesn't break again ;)

PR #145, opened 2015-12-06
LOL LMAO ROFL XD XD XP XP XP

demipixel:
XP

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -45,6 +45,10 @@
       - [server.playNoteBlock(world, position, pitch)](#serverplaynoteblockworld-position-pitch)
       - [server.getNote(note)](#servergetnotenote)
       - [server.emitParticle(particle, world, position, opt)](#serveremitparticleparticle-world-position-opt)
+      - [serv.getXpLevel(xp)](#servgetxplevelxp)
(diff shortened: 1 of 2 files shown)


rom1504:
Can you add doc for these functions ? (so we don't have to add lot and lot of docs later)

rom1504:
Oh well, I kind of failed my merge, but all the commits are in master now.

mhsjlw:
@rom1504, we should really start thinking about a way to store all of this stuff like a function that stores key-values? It would be good to start saving info on players.

rom1504:
@mhsjlw yes, but we probably want to start by using the vanilla storage format

mhsjlw:
Do we really need to do that? I think it would make more sense maybe just making a wrapper around an sqlite3 server that starts when flying-squid does, then just make methods for storing data in the database. Then, if we wanted to we could make a converter that converts the database to world format and so on so forth. Also, we could add abilities to add different databases other than sqlite3 like PostgreSQL and MongoDB and MySQL

rom1504:
@mhsjlw I think we should be able to use different backend, but the default should be vanilla.

But if you want to start working on a non-vanilla backend, feel free to ;)

rom1504:
also see https://github.com/PrismarineJS/prismarine-world/issues/11

PR #148, opened 2015-12-07
Added IP ban command

109C:


rom1504:
Ban shouldn't kick. It works like that usually anyway.
Does vanilla minecraft kick if you ban ip ?

demipixel:
Very likely but I haven't tested

rom1504:
alright I tried, the vanilla server does kick you when you get ip banned

PR #154, opened 2015-12-10
wats ur fav type of icecream

demipixel:
Post in the comments!

### DIFF
--- a/doc/CONTRIBUTE.md
+++ b/doc/CONTRIBUTE.md
@@ -5,93 +5,67 @@
 Directory architecture :
 
 * app.js: specific settings and actually start the server
-* index.js: contain the generic server implementation
-* lib/: contain the classes and functions used in the plugins
-  * serverPlugins/: server plugins that do things general to the server, 
-  properties and method are added to the server object in them
-  * playerPlugins/: player plugins that do things for each player, properties and method are added to the player object in them
+* dist/: Contains "compiled" code (go to current directory in console and type `gulp` to generate this)
+* src/: Source files for the project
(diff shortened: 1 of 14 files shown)


rom1504:
Circle ci points out a few errors, look at it ;)

109C:
Chocolate and vanilla.

doc/CONTRIBUTE.md
```diff
@@ -5,93 +5,67 @@
...
+  player.on(',,,', ...);
+  // serv.on('...', .– don't even think about it
 }
+
 ```
 
 ## Logs and event
 
 In order to keep logging independent from the rest of the server and to let people react in other ways than logging,
-server and player events should be emitting and the logging should only take place in response to these events
```
rom1504: that's still true

rom1504:
when you do `/help command`, the description of the command is displayed 2 times, that's one time too much

demipixel:
will not commit unless u post ur fav ice cream

rom1504:
secondly, why did you remove the usage from the main `/help` ? vanilla has it. (and I think it's useful)

hmm I guess my favorite ice cream might be vanilla, or mango :p

doc/CONTRIBUTE.md
demipixel: serv.log() isn't an event. That's why I added

    logging uses methods and event from `log.js` These include...

rom1504: yes, but what I meant is player and server events are emitted, and then listened to in log.js which call serv.log

PR #161, opened 2015-12-12
[WIP] implement portal detection and use it when flint and steel is used

rom1504:
currently places cactus instead of nether block because nether block seems to require the multi block change packet

lot of tests are already there, but some more need to be added, it seems some cases don't work (when the portal have edges it seems)

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,8 @@
   "dependencies": {
     "babel-runtime": "^5.4.4",
     "emit-then": "^1.0.2",
+    "event-promise": "0.0.1",
+    "flatmap": "0.0.3",
     "minecraft-data": "0.7.0",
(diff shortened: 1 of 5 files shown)


rom1504:
portail frames are now detected when using flint and steel and form the portal.
Missing:
* [ ] remove the portal when breaking a portal block
* [ ] remove the portal when breaking an obsidian
* [ ] store the portal location in a serv.portals
* [ ] create a portal in the nether in the right location
* [ ] when going in the portal, teleport to the corresponding nether portal after 5s

rom1504:
The frame is detected and the portal created. I'm merging this.

The remaining of the portal implem is tracked in https://github.com/mhsjlw/flying-squid/issues/79

PR #165, opened 2015-12-14
Redstone

demipixel:
In progress

### DIFF
--- a/src/lib/plugins/blocks.js
+++ b/src/lib/plugins/blocks.js
@@ -20,7 +20,7 @@ module.exports.player=function(player,serv)
     });
     
 
-  player.setBlock = (position,blockType,blockData) => serv.setBlock(player.world,position,blockType,blockData);
+  player.setBlock = (position,blockType,blockData) => serv.setBlock(player.world,position,blockType,blockData,{blacklist: [player]});
 
 
   player.commands.add({
--- a/src/lib/plugins/blocks.js
+++ b/src/lib/plugins/blocks.js
@@ -4,12 +4,7 @@ module.exports.player=function(player,serv)
 {
   player.changeBlock=async (position,blockType,blockData) =>
   {
-    serv.players
-      .filter(p => p.world==player.world && player!=p)
-      .forEach(p => p.sendBlock(position, blockType, blockData));
-
-    await player.world.setBlockType(position,blockType);
-    await player.world.setBlockData(position,blockData);
+    serv.setBlock(player.world, position, blockType, blockData);
   };
   
   player.sendBlock = (position, blockType, blockData) =>  // Call from player.setBlock unless you want "local" fake blocks
--- a/src/lib/plugins/redstone.js
+++ b/src/lib/plugins/redstone.js
@@ -2,15 +2,14 @@ var Vec3 = require('vec3').Vec3;
 var async = require('async-q');
 
 var redstoneRelated = [55, 75, 76, 93, 94, 123, 124, 152, 356, 404, 29, 33];
+var levelSides = [new Vec3(1, 0, 0,), new Vec3(-1, 0, 0), new Vec3(0, 0, 1), new Vec3(0, 0, -1)]; // Adjacent flat sides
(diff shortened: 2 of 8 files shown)


src/lib/plugins/blocks.js
```diff
@@ -4,12 +4,7 @@ module.exports.player=function(player,serv)
 {
   player.changeBlock=async (position,blockType,blockData) =>
   {
-    serv.players
-      .filter(p => p.world==player.world && player!=p)
```
rom1504: when you dig or place a block you don't need to send that info to the vanilla client, it already knows

109C: @rom1504 The vanilla server echos the block change back to the client, so I think we should too.

demipixel: I don't think it does... Anyway, I already fixed it.

rom1504: indeed the vanilla server doesn't echo it. Except if you (the client) do something wrong

src/lib/plugins/redstone.js
```diff
@@ -0,0 +1,223 @@
...
+        var worldId = worlds.indexOf(d.world);
+        if (!worldId) {
+          worlds.push(d.world);
+          worldId = worlds.length - 1;
+        }
+        posList[d.position.toString() + ',' + worldId] = d;
+      });
+      Object.keys(posList).forEach(pos => {
+        var d = posList[pos];
+        serv.setBlock(d.world, d.position, d.type, d.data);
```
rom1504: setBlock return a promise, better not ignore it

rom1504:
I didn't read the code fully. But I think it would really be a good idea to put all that redstone code in a network-free lib in lib/.
Since it doesn't seem to use the network for the most part.
Only a very small part would then remain in plugins/redstone.js

That would allow for easier testing/optimization. It would also allow to use it outside of flying-squid.

demipixel:
I thought it would be too difficult because of all the getBlocks that there would be.

rom1504:
@demipixel how is that a problem ? you already pass the world anyway
(it's okay if it depends on the world)

demipixel:
Do I have to do that before we speed it up?

rom1504:
No. But it may faster to speed it up if you do that first ;)

rom1504:
I'm afraid https://github.com/mhsjlw/flying-squid/pull/173 made conflicts with this one.
I think you were changing stuff anyway, but you'll have to rebase ;)

mhsjlw:
Can you rebase so we can merge this?

rom1504:
It wasn't ready to merge I think (too slow), but I didn't try it, not sure.
It's related to https://github.com/PrismarineJS/flying-squid/issues/192

mhsjlw:
I'll check it out later.

demipixel:
I don't think this should be merged. It's too old and super laggy.

In other news, I made a [powered wool](https://gfycat.com/GlisteningDemandingBufeo) plugin using ScriptCraft and it seems like it's very quick. Essentially, instead of a redstone dust checking what its power should be (1 less than the surroundings) and then updating everything around it [taking up to 1700 updates for turning off a line of 3 redstone](https://youtu.be/NEMARMNvDsw?t=25s).

The way I fixed it in "powered wool" was this: When a wool enables, it looks at the blocks around it and sets their wool power (instead of wool setting itself). If a wool gets updated and it notices it has an *incorrect* level, it completely turns off. This means turning off a line of 3 takes 3 updates (or 21 checks if we include surrounding blocks). This is because when the lever next to the redstone 15 disables, instead of checking around it and saying "I need to be one less than 14, which is 13!" it says "I need to be one less than 14, but I'm not 13, so I will disable myself!". This is different than how vanilla does it (because blocks update themselves on an update instead of updating others), but it seems like this would be the most efficient way of doing it.

My setup doesn't account for things like powering solid blocks or enabling pistons, droppers, repeaters, or items like that, but the main issue appears to be in dust, so I think this setup would really help.

PR #173, opened 2015-12-17
use const and let instead of var

rom1504:
see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,7 +1,7 @@
-var mc = require('minecraft-protocol');
-var EventEmitter = require('events').EventEmitter;
-var path = require('path');
-var requireIndex = require('requireindex');
+const mc = require('minecraft-protocol');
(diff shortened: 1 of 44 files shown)


rom1504:
Using let instead of var restrict the scope of the variable which reduces the likelihood of having weird behaviors.
Using const when possible has the benefit of let plus clearly mark that the variable is not re-affected. (note this doesn't indicate immutability !)

rom1504:
I only ran the tests on this, I want to make sure there are no problems when actually using the server with a vanilla client. (I wish we had https://github.com/mhsjlw/flying-squid/issues/174)

rom1504:
ok it works

PR #222, opened 2016-03-14
Add limits (Only mobspawn for now)

Creeplays:


rom1504:
> 16:25 < rom1504> it doesn't make a lot of sense
> 16:25 < mhsjlw> @rom1504 ?
> 16:26 < mhsjlw> it doesn't set a global mob limit
> 16:26 < rom1504> you're putting on the same level the max player and the max mobs that can be spawned in a single command
> 16:26 < mhsjlw> it just sets the max for spawning at the same time
> 16:27 < rom1504> yeah but anyway that /summonMany command is just for testing
> 16:27 < rom1504> it'll go in a plugin eventually
> 16:27 < mhsjlw> ok
> 16:27 < rom1504> so it doesn't make much sense to add options for it in settings.json

so maybe we could have a "limit-summon-many" option in the settings.json but it's independant from max-players, max-players should stay as it is.

PR #225, opened 2016-03-20
Block coords!=Entity coords

Creeplays:


rom1504:
Offset doesn't change the position. Test your code ;)

Creeplays:
I broken git on my server, and cannot just commit, and commited tougth web :D

rom1504:
what is this supposed to do ? initEntity should spawn on the given position

Issue #236, opened 2016-04-14
Picking up, throwing, and digging block glitches


javacoolme:
1. Picking up blocks puts them into the corner of your inventory? What?
2. You can't throw blocks with 'q'.
3. When you throw a block out of your inventory, it doesn't drop.
4. Let's say you have three dirt blocks. You place two. It says you have one. Then you dig another, and now you have four dirt blocks. What?
Plz fix

rom1504:
yes inventory is barely implemented.

javacoolme:
Yeah...

109C:
\#4 is literally one line of code... I should be done that soon.

PR #237, opened 2016-04-14
Inventory related stuff for #236

109C:
- [x] Remove block from inventory when placing
- [x] Spawn block when quick dropped


javacoolme:
Awesome!!! Nice job! +1

PR #241, opened 2016-04-30
Added difficulty command

theskiier14:
added the ability to use the difficulty command in the player.js file.

### DIFF
--- a/src/lib/plugins/players.js
+++ b/src/lib/plugins/players.js
@@ -46,7 +46,7 @@ module.exports.player = function(player, serv) {
       return parseInt(str);
     },
     action(diff){
-      player.changeWorld(serv.overworld, diff);
+      serv._writeAll('difficulty',{difficulty:diff});
     }
   });
 };
--- a/src/lib/plugins/players.js
+++ b/src/lib/plugins/players.js
@@ -44,6 +44,7 @@ module.exports.player=function(player){
       return parseInt(str);
     },
     action(diff){
+      serv._writeAll('difficulty', {difficulty: diff});
       serv.difficulty = diff;
     }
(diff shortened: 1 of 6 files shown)


rom1504:
Looks good.
It would be nice to do `serv.difficulty=diff;` and use serv.difficulty instead of 0 there https://github.com/PrismarineJS/flying-squid/search?utf8=%E2%9C%93&q=difficulty&type=Code

And to add a "difficulty" settings in default-settings.json and to use it to define the initial serv.difficulty. (`serv.difficulty=settings.difficulty;` in a server plugin)

src/lib/plugins/players.js
```diff
@@ -30,4 +30,23 @@ module.exports.player=function(player){
       player.setGameMode(mode);
     }
   });
-};
\ No newline at end of file
+};
+
+module.exports.player = function(player, serv) {
```
rom1504: ah you are exporting .player a second time here, that doesn't work, put the command with the gamemode command up there.

src/lib/plugins/players.js
```diff
@@ -30,4 +30,21 @@ module.exports.player=function(player){
...
+    usage: '/difficulty <difficulty>',
+    op: true,
+    parse(str){
+      let results;
+      if(!(results = str.match(/^([0-3])$/)))
+        return false;
+      return parseInt(str);
+    },
+    action(diff){
+      serv.difficulty = diff;
```
rom1504: still need the serv._writeAll too here

rom1504:
can you add back the world/.gitkeep file ?

rom1504:
Nice work !

PR #281, opened 2016-11-20
Added my plugin to the README

DeudlyYT:
I made a plugin that aims to be like the Essentials of Bukkit/Spigot

Actually, its a WIP (work in progress)

I have another suggestion: Why dont create a Discord server (discord servers are free)? It has computer and mobile apps so if people download them and join the server this project could be more live with that chats. Also, it has a rank system


Creeplays:
There is no reason to create this plugin IMHO.
This server is not for normal game, its for e.g minigames servers/e.t.c

demipixel:
@Creeplays I think we should attempt to make the uses as broad as possible. Why limit ourselves?

Creeplays:
Discord is cool, but flying squid developers already have a chat (gitter.im/Prismarine-MC/flying-squid)

rom1504:
Flying-squid can be for normal games.

mhsjlw:
The point of flying-squid is just to be modular, so I guess you can really take it as far as your imagination can take you. I don't really see a reason to limit plugin creation? But a mini game server is a good focus

PR #320, opened 2018-02-06
Migrate to CircleCI 2 and Jest

mhsjlw:
- Reformatted the test files
- `mineflayer.js` is broken
- CircleCI will cache `node_modules/`


mhsjlw:
(Note: `mineflayer.js` is broken)

When the light is green we'll merge.

mhsjlw:
Alright, so I've really gone down the rabbit hole with this one.

The reason why `mineflayer.js` doesn't work is because requireindex loads all of the plugins at startup, it uses `require.extensions` which tells it what modules it is allowed to load. _However_, Jest overrides the default require (presumably to allow hotloading). In doing so, it completely broke requireindex. I have submitted a pull request to fix the problem, but instead, I have just included the slightly modified `requireindex.js` file in the source tree. It is licensed under MIT, and credit is given at the top of the file. Hopefully, we can remove this later and continue using the original requireindex once the bug is fixed.

rom1504:
:+1:

PR #310, opened 2017-08-28
making changes to get 1.12.1 clients working

allain:
Can't hurt to make a pull request to open up the conversation at least.

PR #286, opened 2017-01-11
multiversion support, node8, and 1.9->1.12 support

rom1504:
using https://github.com/PrismarineJS/prismarine-chunk/pull/23 but sadly it doesn't work at all : the player spawns but no chunk is displayed.

Update:
see https://github.com/PrismarineJS/flying-squid/pull/286#issuecomment-310871759 for current status

Issue #219, opened 2016-03-02
1.12.2 Protocol Support


mhsjlw:
See 1.12.2 milestone

rom1504:
first step is at https://github.com/PrismarineJS/prismarine-chunk/issues/19

mhsjlw:
Okey dokey. So here's a list of things that I've found we need to fix so far:

- placeBlock is broken because now in 1.12.2, the client doesn't send us the block with the placeBlock, they actually send us their hand (which we then need to resolve to the item)-- not entirely sure how to add this since we don't really have a way to get this right now (that I know of)
- Sending entity positions to players (i.e.: players can't see each other right now)-- I think this is just some scaling stuff we have to fix (it appears the `/ 32` works for sending the player theirself's position)
- etc.

rom1504:
Yes entities position format has changed

PR #333, opened 2018-05-14
1.12

rom1504:
can connect.

I want to do if(version == "1.12") like in mineflayer because I don't think it's complicated actually.

To do : 
* [x] fix tests : do multiple version tests like nmp and mineflayer
* [ ] check everything
* [x] positions
* [x] send unload chunk in 1.12
* [x] do place block properly (using inventory and not heldItem packet property)
* [x] check https://github.com/PrismarineJS/flying-squid/pull/310 https://github.com/PrismarineJS/flying-squid/pull/286 and https://github.com/PrismarineJS/flying-squid/issues/219
* [x] check http://wiki.vg/Protocol_History and http://wiki.vg/Protocol vs http://wiki.vg/index.php?title=Protocol&oldid=7368


mhsjlw:
FYI when I took a go at this there was an issue with placing or breaking blocks. Some packet no longer has some `hand` field I believe (that gives the item in hand)

rom1504:
We should be concerned by : 
* [ ] Removed "Leash" boolean from Attach Entity - Attach entity is now used only for leashing and Set 
* [ ] Added Set Passengers (0x3A)
* [ ] Added Teleport ID VarInt field to Player Position and Look (0x2E - clientbound)
* [ ] Added Teleport Confirm (0x1D - serverbound) packet, with a single VarInt for the teleport ID
* [ ] Added Sound Event (numeric ID) with ID 0x48. Structure is VarInt for the ID; x, y, and z as ints; volume as float; and pitch as a byte.
* [ ] Added a UUID to Spawn Painting (0x04) after entity ID
* [ ] Added VarInt enum category to Named Sound Effect (0x19)
* [ ] Added VarInt enum category to Sound Effect (0x47)
* [ ] Changed fixed-point ints for x, y, and z coordinates to doubles in Spawn Object (0x00), Spawn Experience Orb (0x01), Spawn Global Entity (0x02), Spawn Mob (0x03), Spawn Player (0x05), and Entity Teleport (0x4A).
* [ ] Changed byte-based fixed deltas measured in 32ndths of a block to short-based fixed deltas measured as (currentX * 32 - prevX * 32) * 128 in Entity Relative Move (0x25) and Entity Look And Relative Move (0x26). As such, these packets are used for movements of at max 8 blocks instead of at max 4 blocks.
* [ ] Changed cursor position values in Player Block Placement (0x1C) from ints ranging from 0 to 15 to floats ranging from 0 to 1.

rom1504:
* [ ] all spawn packet have now uuid : what to put for object and mobs ?

rom1504:
priorities : 
* [ ] positions : 
  * [ ] updatePositions.js
  * [ ] particle.js
  * [ ] entities.js
  * [ ] spawn.js
  * [ ] tp.js
* [ ] delta
  * [ ] physics.js

rom1504:
I'm going to store all the positions as double (new system), and convert back to fixed point for 1.8

rom1504:
Did the position thing.

I'm pretty happy of having finally figured out the solution to store position as double without loss of precision.

See the comment in updatePosition.js for more info.

The remaining things to do here are small stuff.

test/simple.test.js
```diff
@@ -5,33 +5,46 @@ const squid = require('flying-squid')
...
-
-    serv.on('listening', () => {
-      done()
+const { firstVersion, lastVersion } = require('./common/parallel')
+
+squid.supportedVersions.forEach((supportedVersion, i) => {
+  if (!(i >= firstVersion && i <= lastVersion)) {
+    return
+  }
+  const PORT = Math.round(30000 + Math.random() * 20000)
```
mhsjlw: This doesn't smell good

rom1504: why ?
This is what I do in mineflayer and node-minecraft-protocol. It allows for running multiple tests at the same time without any port collision

rom1504:
Tests adapted.
Mineflayer test 1.12 is not passing because of something about prismarine-chunk. See https://circleci.com/gh/PrismarineJS/flying-squid/820#queue-placeholder/containers/1

test/simple.test.js
mhsjlw: You should use the `CIRCLE_NODE_INDEX` environment variable instead, then add that to `3000`, for example.

rom1504: Ah yeah I could do that. It's unlikely to be a problem but why not. I'll change that

rom1504: No, that won't work locally.

mhsjlw: `process.env.CIRCLE_NODE_INDEX || 0`

mhsjlw: Or even `|| (Math.random() * 20000)`

rom1504: fixed, going full deterministic

rom1504: used port 0. Much more reliable. Thanks for insisting on that issue !

rom1504:
left : 
* [ ] attach
* [ ] checking stuff
* [ ] does saving still work ?

mhsjlw:
This features file is really cool, perhaps we should expand it to be admin configuration? So that they can enable/disable features?

rom1504:
Maybe it could use the same pattern but I'd like to keep this one for minecraft version support for now. It's still a bit new and I'm not sure what are all the consequences.
I think there should be a vanillaFeature.json (which is what features.json is currently), and then we can possibly have more feature files if it makes sense for other use cases.

rom1504:
sounds are not converted for 1.12 but let's not do it here (it needs more data and stuff), tracked there https://github.com/PrismarineJS/flying-squid/issues/340

rom1504:
works nicely. Let's merge

PR #354, opened 2018-07-09
Fix #353 -- Helditem issue fixed

### LINKED ISSUE #353: Helditem was being checked even if there was no held item
This causes the server to crash whenever you tried to interact with an object with nothing in the hand

Deudly:
More info at the issue
Ref #353


rom1504:
This is breaking the tests, please fix ;)

rom1504:
fixed by other pr

PR #358, opened 2019-01-08
Fixed #353 - Crash Helditem

### LINKED ISSUE #353: Helditem was being checked even if there was no held item

bitknox:
Ref #353

### DIFF
--- a/src/lib/plugins/placeBlock.js
+++ b/src/lib/plugins/placeBlock.js
@@ -15,7 +15,7 @@ module.exports.player = function (player, serv, { version }) {
 
   player._client.on('block_place', ({ direction, location } = {}) => {
     const heldItem = player.inventory.slots[36 + player.heldItemSlot]
-    if(heldItem == undefined) return
+    if (heldItem === undefined) return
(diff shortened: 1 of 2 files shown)


src/lib/plugins/placeBlock.js
```diff
@@ -15,6 +15,7 @@ module.exports.player = function (player, serv, { version }) {
 
   player._client.on('block_place', ({ direction, location } = {}) => {
     const heldItem = player.inventory.slots[36 + player.heldItemSlot]
+    if(heldItem == undefined) return
```
rom1504: use ===

src/lib/plugins/useItem.js
```diff
@@ -5,6 +5,7 @@ module.exports.player = function (player, serv, { version }) {
 
   player._client.on('block_place', ({ direction, location } = {}) => {
     const heldItem = player.inventory.slots[36 + player.heldItemSlot]
+    if(heldItem == undefined) return
```
rom1504: use ===
and add a space before the ( 
(or run standard --fix to fix these kind of things)

PR #359, opened 2019-01-25
[FIX] import all plugins and add script on package.json

Himself65:


Himself65:
What? Why didn't pass the test

rom1504:
it's already done at https://github.com/PrismarineJS/flying-squid/blob/e28e5b0e6285eed30720c214120d021ac8f52b91/src/lib/plugins/external.js

Himself65:
@rom1504 nice work

PR #361, opened 2019-01-27
fix setblock command typo

Himself65:
I think this can as the example and also test project


test/mineflayer.test.js
```diff
@@ -243,11 +243,12 @@ squid.supportedVersions.forEach((supportedVersion, i) => {
         await waitMessage(bot, 'You do not have permission to use this command')
         serv.getPlayer('bot').op = true
       })
-      test('can use /setblock', async () => {
+      test('can use /setblock', async (done) => {
         await Promise.all([waitSpawnZone(bot, 2), onGround(bot)])
         bot.chat('/setblock 1 2 3 95 0')
         let [, newBlock] = await once(bot, 'blockUpdate:' + new Vec3(1, 2, 3), { array: true })
         expect(newBlock.type).toEqual(95)
+        done()
```
Himself65: I don't know why this part always throws an error which display it time out. I don't think this part have any problem before.

Himself65:
@rom1504 maybe i fixed the bug on setblock？

Himself65:
and you need squash these commits when merging

src/lib/plugins/blocks.js
```diff
@@ -11,7 +11,7 @@ module.exports.player = function (player, serv) {
   }
 
   player.sendBlock = (position, blockType, blockData) => // Call from player.setBlock unless you want "local" fake blocks
-    player.behavior('sendBlock', {
+    player.behavior('sendblock', {
```
rom1504: Why did you change that, was it wrong ? Are the docs up to date on this ?

test/plugin.test.js
```diff
@@ -0,0 +1,45 @@
...
+const settings = require('../config/default-settings')
+const path = require('path')
+const fs = require('fs')
+
+describe('test import third-party plugins', () => {
+  const pluginPath = path.resolve(__dirname, '../', 'src', 'plugins', 'example')
+  const filePath = path.resolve(__dirname, '../', 'src', 'plugins', 'example', 'index.js')
+  beforeAll(() => {
+    if (!fs.existsSync(pluginPath)) {
+      fs.mkdirSync(pluginPath)
```
rom1504: Why are you writing the file here if you also commit it ?

src/lib/plugins/blocks.js
rom1504: Ok then it should be updated the other way round.
The correct spelling is sendBlock, see https://github.com/PrismarineJS/flying-squid/blob/master/docs/API.md#sendblock

test/plugin.test.js
Himself65: After thinking twice, I think this unit test part is unuseful, so I removed it

src/lib/plugins/blocks.js
Himself65: Fixed

demipixel: Oh wait, the command "setblock" shouldn't be changed...

rom1504:
the vanilla command is `setblock`, see https://minecraft.gamepedia.com/Commands/setblock
let's keep the same name

PR #363, opened 2019-04-16
Fixed mob eggs spawning

rtm516:


rom1504:
Can someone check whether that works ? seems pretty weird to me (cc @Karang maybe)

Karang:
> Can someone check whether that works ? seems pretty weird to me (cc @Karang maybe)

This PR seems to fix the mob eggs spawning crash. With this fix, the correct mob is spawning, without the server is crashing. The physic of the entities is weird but I guess that's out of the scope of this PR.

rom1504:
ok let's merge then

PR #367, opened 2019-07-21
Update header.js

GoogleSites:


rom1504:
That's wrong, fix your config

mhsjlw:
Including the header is documented in the README here: https://github.com/PrismarineJS/flying-squid/blob/master/docs/README.md#L92-L93, but I believe it would make sense to squash this once and for all by updating the `basic.js` example with the one in the README (see: https://github.com/PrismarineJS/flying-squid/pull/368)

PR #368, opened 2019-07-21
Update basic.js

GoogleSites:


mhsjlw:
I personally think it would make more sense if we just synchronize `basic.js` with the example provided in the README (https://github.com/PrismarineJS/flying-squid/blob/master/docs/README.md#L68-L98), which would be a welcome change

PR #378, opened 2020-04-01
Add some vanilla behavior

jvyden420:
Fixed: /time set will now check for invalid numbers.
Fixed: Capped the velocity command to prevent getting kicked
Fixed: Changed the "No entity named" error from a chat message to an error.

Added: Plants will now "break" when the block under it has been broken. (I'm not sure if the way i'm doing it is correct)
Added: /tp will now teleport you to the center of a block when a decimal is not specified.
Added: Some basic tab complete functionality

lmk if i did something horribly wrong

### DIFF
--- a/src/lib/plugins/chat.js
+++ b/src/lib/plugins/chat.js
@@ -157,4 +157,12 @@ module.exports.player = function (player, serv) {
     if (typeof message === 'string') message = serv.parseClassic(message)
     player._client.write('chat', { message: JSON.stringify(message), position: 2 })
   }
+
+  player._client.on('tab_complete', function (data) {
+    let playerNames = []
+    for (let player of serv.players) playerNames.push(player.username)
+    player._client.write('tab_complete', {
+      matches: playerNames
+    })
+  })
 }
--- a/src/lib/plugins/daycycle.js
+++ b/src/lib/plugins/daycycle.js
@@ -52,16 +52,20 @@ module.exports.player = function (player, serv) {
       if (action === 'query') {
         player.chat('It is ' + serv.time)
       } else {
-        let newTime
+        if (isNaN(value)) {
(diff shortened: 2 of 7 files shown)


rom1504:
You broke the tests apparently. Could you fix them ?

jvyden420:
i always forget to run tests before creating a pr, lol
lemme fix them real quick

jvyden420:
did circleci break? i'm not getting these errors on my machine

jvyden420:
i ran the tests that failed on my machine and they seem to fail at random, so i dont think _i_ broke anything

rom1504:
You can see the circle ci links on GitHub for your pull request. They are passing for all the commits in master.

Can you check that link and look what is wrong ?

jvyden420:
I've been running tests for a while and they do indeed seem to randomly pass and fail. Then I remembered the _one_ thing thats unique between each run of the tests: the random seed.

I added some code to manually set the seed to one that passes, and sure enough, the test passed 100% of the time. Similarly, if I set the seed to one that fails, that will fail 100% of the time. Do you mind if I manually specify a seed to make the tests consistent?

This random failing behavior still seems to happen when pulling directly from master, so my code isn't at fault here.You can see it happening [here](https://pastebin.com/3PRuqTv3).

and thats my phoenix wright for the day.

rom1504:
Ok. Fixing the seed seems like a ok workaround.
It would be much better to be able to actually fix the test though but that can be for another time

jvyden420:
allllrighty, this next push should fix it then

rom1504:
LGTM, thanks

do you plan to work some more on flying squid ?

jvyden420:
that's gonna be a definite yeah, this project is pretty cool and I don't think I've seen it done before

Issue #166, opened 2015-12-14
loading/saving


rom1504:
This will probably happen mostly in prismarine-world, but opening an issue here too to track progress.

Player file format is http://minecraft.gamepedia.com/Player.dat_format
which is actually pretty simple (just got to use prismarine-nbt to read/write it)

@mhsjlw is working on an alternative save format using sql (which will save the same things as vanilla)

(we will support the vanilla format and the api will work with any other format, the default can be chosen later)

rom1504:
see https://github.com/PrismarineJS/prismarine-world/issues/7

demipixel:
We need to save:
- Inventories
- Worlds
- Entity NBT data
- Plugin data
  - Could be things like any of the above OR server/world data (server data being options, player data being their plugin attack level, coins they have, etc)

rom1504:
This is mostly done

Other issues can be opened for specifics

PR #380, opened 2020-04-18
Add support for reading <player>.dat files

rrwr:


rrwr:
I think this is related to #166

rom1504:
not sure if this should be happening here vs prismarine-world

rom1504:
I guess let's have it here for now

PR #381, opened 2020-05-01
Add block update queue (+ simple redstone)

Karang:
This PR adds a block update queue. Vanilla's block updates are handled via recursive function call, but this has led to exploits like block update suppression using stack overflows. Instead, the update queue aims at processing a certain amount of block updates per tick and delays remaining updates to the next ticks if too many are coming at the same time. The queue is ordered by tick to ensure correct behavior, the block updates can get behind server tick but never ahead. One other advantage of using a queue instead of function calls is the ability to remove redundant updates by checking if they are already in the queue.

Block update are customized by block types, by registering an update Handler function:
```javascript
serv.onBlockUpdate('redstone_wire', async (world, block, tick) => {
  // ...
}
```
As an example, I provided a simple redstone implementation (only torchs and wires).

This PR also adds a way to customize block placement (for block coming from an item or blocks with orientation). Example:
```javascript
serv.onItemPlace('redstone_torch', (item, direction) => {
  const directionToData = [0, 5, 4, 3, 2, 1]
  return {id: redstone_torch_type, data: directionToData[direction]}
})
```

This is a WIP because the block update mecanisme sill needs some improvements (TODOs in the code) and the redstone implementation still has some difference with vanilla's behavior.


rom1504:
I like it but do you think some of that could go to https://github.com/PrismarineJS/prismarine-world ?

the point of prismarine-world is to provide a world implementation that can be used in flying-squid, mineflayer but also other stuff

if possible, things that are only related to the world (and do not interact directly with packets), should go there

Karang:
> I like it but do you think some of that could go to https://github.com/PrismarineJS/prismarine-world ?
> 
> the point of prismarine-world is to provide a world implementation that can be used in flying-squid, mineflayer but also other stuff
> 
> if possible, things that are only related to the world (and do not interact directly with packets), should go there

The block updates also interact with players and entities. It needs to send packet update to players for them to see the changes. It also needs to listen to the server's tick event. If we want to put some part in prismarine-world, the code handling block updates would have to be split. The way I see it is that blockUpdates.js is just another plugin with the core mechanics, then we have other plugins per block types (eg: redstone.js).

Block update simulation is something that the client does not need to do. Mineflayer might use some redstone simulator to plan ahead, but the way to simulate it would have different requirements than the server's simulator

rom1504:
Ok let's keep it in flying-squid

Karang:
This WIP is mostly done. There is still the locking feature of the repeaters to be implemented and maybe try to factorize a bit the logic of redstone.js, but I believe the API of block update will not change.

Here is a small documentation of the API for easy review:

### Block placement customization

Some blocks have a different id than the item they are spawned from or have a metadata dependent on the placement context. These blocks can be customized by registering an itemPlace handler on the held item type (one handler per item type). For instance, the plugin that customize the signs placement:
```javascript
module.exports.server = (serv, { version }) => {
  const mcData = require('minecraft-data')(version)

  const oakSignType = mcData.blocksByName.standing_sign.id
  const oakWallSignType = mcData.blocksByName.wall_sign.id

  serv.on('asap', () => {
    serv.onItemPlace('sign', ({ item, direction, angle }) => {
      if (direction === 1) {
        return { id: oakSignType, data: Math.floor(angle / 22.5 + 0.5) & 0xF }
      }
      return { id: oakWallSignType, data: direction }
    })
  })
}
```
The argument given to the handler is an object containing the held item that triggered the event, the direction (face) on which the player clicked, the angle of the player around the placed block. It should return an object containing the id and data of the block to place.

### Block interaction

This handler is called when a player interact with a block.
```javascript
module.exports.server = (serv, { version }) => {
  serv.on('asap', () => {
    const repeaterInteraction = async ({ block, player }) => {
      const data = (block.metadata + 4) & 0xF
      player.setBlock(block.position, block.type, data)
      return true
    }
    serv.onBlockInteraction('powered_repeater', repeaterInteraction)
    serv.onBlockInteraction('unpowered_repeater', repeaterInteraction)
  }
}
```
The argument given to the handler is an object containing the clicked block and the player. It should return true if the block interaction occurred and the block placement should be cancelled.

### Block update

This handler is called when a block of the given type is updated. It should verify that the block state is still correct according to the game's rules. It is triggered when a neighboring block has been modified.
```javascript
module.exports.server = (serv, { version }) => {
  serv.on('asap', () => {
    serv.onBlockUpdate('redstone_wire', async (world, block, fromTick, tick, data) => {
      const pos = block.position

      // Redstone wire should be on a solid block
      const support = await world.getBlock(pos.offset(0, -1, 0))
      if (support.boundingBox !== 'block') {
        await world.setBlockStateId(pos, 0)
        serv.notifyNeighborsOfStateChange(world, pos, tick, tick)
        return true
      }
      // ... rest of redstone logic
      return changed
    })
  }
}
```
The arguments of the handler are the world in which the update occurred, the block, fromTick the tick at which the update was triggered, the tick the update was scheduled to (current tick), and optional data (null most of the time) that can be used to transmit data between block updates. The handler should return true if the block was changed so the update manager can send a multiBlockChange packet for all the changes that occurred within the tick. The state of the block should be modified by using the world's setBlockXXX functions instead of serv.setBlock (that would send redundant updates to players).

rom1504:
Could you put these explanations in a `block-update.md` in https://github.com/PrismarineJS/flying-squid/tree/master/docs ?

Karang:
> Could you put these explanations in a `block-update.md` in https://github.com/PrismarineJS/flying-squid/tree/master/docs ?

done

I think everything is here for this PR. I'm removing the WIP. If its ok for you this is ready for merging.

rom1504:
can you add the new methods you exposed to https://github.com/PrismarineJS/flying-squid/blob/master/docs/API.md ?

Karang:
> can you add the new methods you exposed to https://github.com/PrismarineJS/flying-squid/blob/master/docs/API.md ?

done

rom1504:
thanks

rom1504:
btw @Karang out of curiosity, do you have any particular bigger plan for using this / this king of flying squid improvements ?

Karang:
> btw @Karang out of curiosity, do you have any particular bigger plan for using this / this king of flying squid improvements ?

I'm interested to see how far we can push node to be a high performance minecraft server. Optimizing code is my research area, but here I'm doing it for fun :)

rom1504:
Ok cool :)

PR #383, opened 2020-05-03
Add support for block actions

FalcoG:
This will allow a plugin to make use of block actions. They are required to visually open chests, pistons and more.

wiki.vg reference: https://wiki.vg/Block_Actions
minecraft-data reference: https://minecraft-data.prismarine.js.org/?d=protocol#toClient_block_action


rom1504:
how did you check whether it work ?
a good solution might be to add a command for this and then a test in the mineflayer test

FalcoG:
I've tested it with the plugin that I wrote, you can view it [here](https://gitlab.com/fgerritsjans/obsidian-ender-chest/-/blob/feature/set-block-action/index.js).

>  a good solution might be to add a command for this and then a test in the mineflayer test

Sounds like a good idea. I'll try adding it to the mineflayer test.

rom1504:
thanks

Issue #394, opened 2020-06-16
Server is not starting


redcarti:
Hi! I was trying to start the server via cloning the repository. 
So, I installed all packages, runned ``npm run prepare``. And when I was trying to connect, flying-squid throwed an error ``RangeError [ERR_OUT_OF_RANGE]: The value of "value" is out of range. It must be >= 0 and <= 4294967295. Received -1``.

redcarti:
All stack trace:
```A:\flying-squid\src\lib\plugins\world.js:177
      .catch((err) => setTimeout(() => { throw err }), 0)
                                         ^

RangeError [ERR_OUT_OF_RANGE]: The value of "value" is out of range. It must be >= 0 and <= 4294967295. Received -1
    at checkInt (internal/buffer.js:58:11)
    at writeU_Int32BE (internal/buffer.js:788:3)
    at Buffer.writeUInt32BE (internal/buffer.js:801:10)
    at SmartBuffer._writeNumberValue (A:\flying-squid\node_modules\smart-buffer\build\smartbuffer.js:1220:14)
    at SmartBuffer.writeUInt32BE (A:\flying-squid\node_modules\smart-buffer\build\smartbuffer.js:467:21)
    at BitArray.writeBuffer (A:\flying-squid\node_modules\prismarine-chunk\src\pc\common\BitArray.js:117:19)
    at ChunkSection.write (A:\flying-squid\node_modules\prismarine-chunk\src\pc\1.9\ChunkSection.js:191:19)
    at A:\flying-squid\node_modules\prismarine-chunk\src\pc\1.9\ChunkColumn.js:165:19
    at Array.forEach (<anonymous>)
    at ChunkColumn.dump (A:\flying-squid\node_modules\prismarine-chunk\src\pc\1.9\ChunkColumn.js:163:21)
    at A:\flying-squid\src\lib\plugins\world.js:126:26
    at Entity.behavior (A:\flying-squid\src\lib\behavior.js:24:14)
    at async Entity.player.login (A:\flying-squid\src\lib\plugins\login.js:218:5)
    at async Server.<anonymous> (A:\flying-squid\src\lib\plugins\login.js:33:7) {
  code: 'ERR_OUT_OF_RANGE'
}```

rom1504:
Looks like a recent update I did for prismarine-chunk caught a bug in flying-squid. Apparently we set the light to -1 somewhere in flying-squid which is incorrect.
I'll try to pinpoint and fix this

[...]

PR #395, opened 2020-06-16
use pchunk branch to try and fix chunks

rom1504:
pushed as example, but it's still failing
trying to fix https://github.com/PrismarineJS/flying-squid/issues/394


rom1504:
no error with this but world is empty from client PoV

Issue #401, opened 2020-06-17
Run MC cmds via terminal


redcarti:
Java servers can run cmds via terminal input. I have tried to implement this but there's no method to invoke cmd. Can this be implemented in flying squid?

rom1504:
Yes

redcarti:
So, how can I implement invoking a cmd?

rom1504:
You can use repl.js module and for example https://github.com/PrismarineJS/flying-squid/blob/master/docs/API.md#playerhandlecommandcommand

redcarti:
Oh, I see. Well, will try tomorrow.

redcarti:
Method ``player.handleCommand`` handles command at player side, not from server side

rom1504:
It just means it has a player context, it's all server side.
You could either run the command as a player or define a way to run
commands as not a player but then you'll for example won't be able to do
set block with relative position.
What does vanilla do about this ?

redcarti:
I don't know how to explain this.. But, I launched second MC, and tried to call ``gamemode 1`` & ``gamemode 0`` commands. So, all players had creative and then survival gamemode.

redcarti:
I think that java servers handles commands from terminal input, e.g ``gamemode 1 [player]`` on one player

rom1504:
Ok, that's something that can be implemented

redcarti:
So, code of this plugin (``stdinLineByLine`` found on Stack Overflow)

``js
const EventEmitter = require('events');

function stdinLineByLine() {
  const stdin = new EventEmitter();
  let buff = "";

  process.stdin
    .on('data', data => {
      buff += data;
      lines = buff.split(/[\r\n|\n]/);
      buff = lines.pop();
      lines.forEach(line => stdin.emit('line', line));
    })
    .on('end', () => {
      if (buff.length > 0) stdin.emit('line', buff);
    });

  return stdin;
}

const stdin = stdinLineByLine();

module.exports.server = function (serv, settings) {

}

module.exports.player = function (player, settings) {
    stdin.on('line', (data) => {
        player.handleCommand(data)
        .then(d => {
            console.log(d)
        })
    });
}
``

rom1504:
Ok, does that work for you ?

redcarti:
It works but I said that it works on everybody.

rom1504:
Ok, you can implement the command /gamemode <gamemode> <player> or adapt the existing one then

redcarti:
[...]

PR #402, opened 2020-06-18
Run MC cmds via terminal & serv.commands

redcarti:
Pull request of #401


redcarti:
It needs to be in other files but whatever..

redcarti:
And it needs to be rewrited... cuz 
image

redcarti:
hmm, thats strange

redcarti:
@rom1504 are throwing an error is needed? I think that just sending message is better.
Also, about ``server.selectorString``, why do it need ``pos``? E.g. ``/kill @a`` don't needs the ``pos``.

redcarti:
Oh, I see that ``handleCommand`` checks ``UserError``. Whatever, will change.

redcarti:
Needed some more cmds..

redcarti:
Input needs to be at the end of console. But that's a problem...

redcarti:
Alright, I tried many different ways to make ``process.stdin`` cursor to the end of a console. But I didn't find best way to do this.

redcarti:
I think that's all for this feature. It can be upgraded.

redcarti:
@rom1504 so?

redcarti:
I think we could add to cmds ``player.chat`` thing, e.g. when executed ``/op <player>``, player that opped will recieve _[<opplayer>: Opped <player>]_ or sth. It's implemented in my ``input.js`` plugin.

rom1504:
Looks good, I'll review it when I have some time

redcarti:
Okay, I'll wait. About last comment, I'll make it soon and create pull request.

rom1504:
Ok reviewed.
Overall it looks good but there is one point that I think would improve this whole feature.

Instead of duplicating all the commands between player and server, what about defining them once and putting them both on the player and on the server ?
If you look at https://github.com/PrismarineJS/flying-squid/pull/402/files#diff-3d81549536b6e81c8b52be79354cdf85R236 vs https://github.com/PrismarineJS/flying-squid/blob/43b665bb84afc44f58758671d5a1e8bc75809cbe/src/lib/plugins/moderation.js#L188 for example, it's exactly the same thing.

The only difference is that in the second case, player is coming from the context, whereas in the first case player is found from the argument.
It should be possible to define it all once and use it to do serv.commands.add and player.commands.add (or even better, make a single .add that will do both)

Do you think you could try doing this ?

I think this is the right time to do this, before everything is duplicated and hard to merge

redcarti:
I think, that might be possible, adding the feature of context in ``Command`` class. Or you meant something different?

redcarti:
I think ``serv.commands`` instead of ``player.commands`` might be better.

redcarti:
And, ``player.handleCommand`` will be using ``serv.commands.use()``

redcarti:
And yeah, so I'm in PrismarineJS discord, so discussion about this feature might be faster

redcarti:
So,
```js
player.handleCommand = async (str) => {
    try {
      const res = await serv.commands.use(str, player)
      if (res) player.chat(serv.color.red + res)
    } catch (err) {
      if (err.userError) player.chat(serv.color.red + 'Error: ' + err.message)
      else setTimeout(() => { throw err }, 0)
    }
  }

serv.handleCommand = async (str) => {
    try {
      const res = await serv.commands.use(str)
      if (res) serv.log('[INFO]: ' + res)
    } catch (err) {
      if (err.userError) serv.log('[ERR]: ' + err.message)
      else setTimeout(() => { throw err }, 0)
    }
  }
```

src/lib/plugins/commands.js
```diff
@@ -1,81 +1,105 @@
...
+  serv.handleCommand = async (str) => {
+    try {
+      const res = await serv.commands.use(str)
+      if (res) serv.info(res)
+    } catch (err) {
+      if (err.userError) serv.err(err.message)
+      else setTimeout(() => { throw err }, 0)
+    }
+  }
+  /*
```
rom1504, line 30: Why is this commented ?

src/lib/plugins/pvp.js
```diff
@@ -30,7 +30,8 @@ module.exports.player = function (player, serv) {
     if (mouse === 1) { attackEntity(target) }
   })
 
-  player.commands.add({
+  /*
```
rom1504, line 33: Why commented ?

src/lib/plugins/blocks.js
```diff
@@ -60,14 +62,14 @@ module.exports.player = function (player, serv) {
       if (!results) return false
       return results
     },
-    action (params) {
+    action (params, ctx) {
       let res = params.slice(1, 4)
-      res = res.map((val, i) => serv.posFromString(val, player.position[['x', 'y', 'z'][i]]))
-      player.setBlock(new Vec3(res[0], res[1], res[2]).floored(), params[4], params[5] || 0)
+      res = res.map((val, i) => serv.posFromString(val, ctx.player.position[['x', 'y', 'z'][i]]))
```
rom1504, line 67: You need to either do something appropriate if ctx.player is undefined or mark this as playerOnly

Same for all the commands

src/lib/plugins/commands.js
redcarti: forgot to delete this

src/lib/plugins/pvp.js
redcarti: as gm, forgot to delete

src/lib/plugins/blocks.js
redcarti: will work on it soon

src/lib/plugins/blocks.js
```diff
@@ -60,14 +62,16 @@ module.exports.player = function (player, serv) {
       if (!results) return false
       return results
     },
-    action (params) {
+    action (params, ctx) {
       let res = params.slice(1, 4)
-      res = res.map((val, i) => serv.posFromString(val, player.position[['x', 'y', 'z'][i]]))
-      player.setBlock(new Vec3(res[0], res[1], res[2]).floored(), params[4], params[5] || 0)
+      if (ctx.player) res = res.map((val, i) => serv.posFromString(val, ctx.player.position[['x', 'y', 'z'][i]]))
+      else res = res.map((val, i) => serv.posFromString(val, new Vec3(0, 128, 0)[['x', 'y', 'z'][i]]))
```
rom1504, line 68: Why this position ? Does the vanilla server do that ?

src/lib/plugins/commands.js
```diff
@@ -1,81 +1,59 @@
...
-  player.commands.add({
+  serv.commands.add({
     base: 'modpe',
     info: 'for modpe commands',
     usage: '/modpe <params>',
     parse (str) { return str || false },
-    action (str) {
-      player.emit('modpe', str)
+    action (str, ctx) {
+      ctx.player.emit('modpe', str)
```
rom1504, line 52: Handle ctx.player undefined or mark as player only

src/lib/plugins/commands.js
```diff
@@ -93,36 +71,102 @@ module.exports.player = function (player, serv, { version }) {
...
     usage: '/selector <selector>',
     op: true,
     parse (str) {
       return str || false
     },
-    action (sel) {
-      const arr = serv.selectorString(sel, player.position, player.world)
-      player.chat(JSON.stringify(arr.map(a => a.id)))
+    action (sel, ctx) {
+      const arr = serv.selectorString(sel, ctx.player.position, ctx.player.world)
```
rom1504, line 83: Handle ctx.player undefined

src/lib/plugins/daycycle.js
```diff
@@ -48,9 +46,9 @@ module.exports.player = function (player, serv) {
         value: data[2] === 'day' ? 1000 : (data[2] === 'night' ? 13000 : parseInt(data[2]))
       }
     },
-    action ({ action, value }) {
+    action ({ action, value }, ctx) {
       if (action === 'query') {
-        player.chat('It is ' + serv.time)
+        ctx.player.chat('It is ' + serv.time)
```
rom1504, line 51: Handle ctx.player undefined

src/lib/plugins/effects.js
```diff
@@ -50,16 +50,16 @@ module.exports.entity = function (entity, serv) {
...
     base: 'effect',
     info: 'Give player an effect',
     usage: '/effect <player> <effect> [seconds] [amplifier] [hideParticles]',
     parse (str) {
       return str.match(/(.+?) (\d+)(?: (\d+))?(?: (\d+))?(?: (true|false))?|.*? clear/) || false
     },
-    action (params) {
-      const targets = player.selectorString(params[1])
+    action (params, ctx) {
+      const targets = ctx.player.selectorString(params[1])
```
rom1504, line 62: Handle ctx.player undefined

src/lib/plugins/effects.js
```diff
@@ -78,9 +78,9 @@ module.exports.player = function (player) {
         })
       }
       const chatSelect = (targets.length === 1 ? (targets[0].type === 'player' ? targets[0].username : 'entity') : 'entities')
-      if (params[2] === 'clear') player.chat('Remove all effects from ' + chatSelect + '.')
+      if (params[2] === 'clear') ctx.player.chat('Remove all effects from ' + chatSelect + '.')
```
rom1504, line 81: Handle ctx.player undefined

src/lib/plugins/experience.js
```diff
@@ -29,30 +29,32 @@ module.exports.player = function (player, serv) {
...
       const amt = parseInt(args[1])
-      const user = args[3] ? serv.getPlayer(args[3]) : player
+      const user = args[3] ? serv.getPlayer(args[3]) : ctx.player
       if (!user) return args[3] + ' is not on this server!'
 
       if (!isLevel) {
         user.setXp(user.xp + amt)
-        player.chat('Gave ' + user.username + ' ' + amt + ' xp')
+        ctx.player.chat('Gave ' + user.username + ' ' + amt + ' xp')
       } else {
```
rom1504, line 52: Handle ctx.player undefined

redcarti:
idk, if I type ``setblock ~1 ~1 ~1 stone``, server will return ``Cannot place block outside of the world``

src/lib/plugins/moderation.js
```diff
@@ -89,18 +137,18 @@ module.exports.player = function (player, serv) {
         reason: parts.join(' ')
       }
     },
-    action ({ username, reason }) {
+    action ({ username, reason }, ctx) {
       const kickPlayer = serv.getPlayer(username)
       if (!kickPlayer) {
-        player.chat(username + ' is not on this server!')
+        ctx.player.chat(username + ' is not on this server!')
```
rom1504, line 143: Handle ctx.player undefined

src/lib/plugins/moderation.js
```diff
@@ -113,28 +161,28 @@ module.exports.player = function (player, serv) {
...
+    action ({ username, reason }, ctx) {
       const banPlayer = serv.getPlayer(username)
 
       if (!banPlayer) {
         serv.banUsername(username, reason)
           .then(() => {
-            serv.emit('banned', player, username, reason)
-            player.chat(username + ' was banned')
+            serv.emit('banned', ctx.player, username, reason)
+            ctx.player.chat(username + ' was banned')
```
rom1504, line 171: Handle ctx.player undefined

src/lib/plugins/moderation.js
```diff
@@ -148,24 +196,24 @@ module.exports.player = function (player, serv) {
...
+  serv.commands.add({
     base: 'pardon-ip',
     info: 'to pardon a player by ip',
     usage: '/pardon-ip <ip>',
     op: true,
-    action (IP) {
+    action (IP, ctx) {
       const result = serv.pardonIP(IP)
-      player.chat(result ? IP + ' was IP pardoned' : IP + ' is not banned')
+      ctx.player.chat(result ? IP + ' was IP pardoned' : IP + ' is not banned')
```
rom1504, line 212: Handle ctx.player undefined

redcarti:
So, for a while it might be ``0, 128, 0``

redcarti:
Okay, workin on it

src/lib/plugins/blocks.js
redcarti: idk, if I type ``setblock ~1 ~1 ~1 stone``, server will return ``Cannot place block outside of the world``

redcarti:
wait

src/lib/plugins/commands.js
redcarti: ``onlyPlayer: true``

src/lib/plugins/commands.js
```diff
@@ -93,36 +72,103 @@ module.exports.player = function (player, serv, { version }) {
...
 
-  player.handleCommand = async (str) => {
-    try {
-      const res = await player.commands.use(str, player.op)
-      if (res) player.chat(serv.color.red + res)
-    } catch (err) {
-      if (err.userError) player.chat(serv.color.red + 'Error: ' + err.message)
-      else setTimeout(() => { throw err }, 0)
+  serv.commands.add({
+    base: 'help',
```
rom1504, line 91: Could you exclude only player from server help and only console from player help ?

src/lib/plugins/experience.js
```diff
@@ -29,30 +29,34 @@ module.exports.player = function (player, serv) {
...
         user.setXp(user.xp + amt)
-        player.chat('Gave ' + user.username + ' ' + amt + ' xp')
+        if(ctx.player) ctx.player.chat('Gave ' + user.username + ' ' + amt + ' xp')
+        else serv.log('Gave ' + user.username + ' ' + amt + ' xp')
       } else {
-        const currLevel = getXpLevel(player.xp)
+        const currLevel = getXpLevel(ctx.player.xp)
         const baseCurrLevel = getBaseXpFromLevel(currLevel)
-        const extraXp = player.xp - baseCurrLevel
+        const extraXp = ctx.player.xp - baseCurrLevel
```
rom1504, line 56: Check this one too

src/lib/plugins/moderation.js
```diff
@@ -89,18 +137,19 @@ module.exports.player = function (player, serv) {
...
+    action ({ username, reason }, ctx) {
       const kickPlayer = serv.getPlayer(username)
       if (!kickPlayer) {
-        player.chat(username + ' is not on this server!')
+        if(ctx.player) ctx.player.chat(username + ' is not on this server!')
+        else throw new UserError(username + ' is not on this server!')
       } else {
         kickPlayer.kick(reason)
-        kickPlayer.emit('kicked', player, reason)
+        kickPlayer.emit('kicked', ctx.player, reason)
```
rom1504, line 147: Need to check this one

src/lib/plugins/moderation.js
```diff
@@ -148,24 +199,25 @@ module.exports.player = function (player, serv) {
         reason: argv.shift()
       }
     },
-    action ({ IP, reason }) {
+    action ({ IP, reason }, ctx) {
       serv.banIP(IP, reason)
-      player.chat('' + IP + ' was IP banned')
+      ctx.player.chat('' + IP + ' was IP banned')
```
rom1504, line 204: Handle undefined ctx player

src/lib/plugins/moderation.js
```diff
@@ -174,48 +226,33 @@ module.exports.player = function (player, serv) {
       if (!str.match(/([a-zA-Z0-9_]+)/)) { return false }
       return str
     },
-    action (nick) {
+    action (nick, ctx) {
       serv.pardonUsername(nick)
-        .then(() => player.chat(nick + ' is unbanned'))
+        .then(() => ctx.player.chat(nick + ' is unbanned'))
```
rom1504, line 231: Handle this one

src/lib/plugins/particle.js
```diff
@@ -39,13 +37,13 @@ module.exports.player = function (player, serv) {
         size: results[5] ? new Vec3(parseInt(results[3]), parseInt(results[4]), parseInt(results[5])) : new Vec3(1, 1, 1)
       }
     },
-    action ({ particle, amount, size }) {
+    action ({ particle, amount, size }, ctx) {
       if (amount >= 100000) {
-        player.chat('You cannot emit more than 100,000 particles!')
+        ctx.player.chat('You cannot emit more than 100,000 particles!')
         return
```
rom1504, line 43: Handle this

src/lib/plugins/physics.js
```diff
@@ -96,17 +96,17 @@ module.exports.entity = function (entity, serv, { version }) {
...
     info: 'Push velocity on player(s)',
     usage: '/velocity <player> <x> <y> <z>',
     op: true,
     parse (str) {
       return str.match(/(.+?) (\d+) (\d+) (\d+)/) || false
     },
-    action (params) {
-      const selector = player.selectorString(params[1])
+    action (params, ctx) {
+      const selector = ctx.player.selectorString(params[1])
```
rom1504, line 109: Handle this

src/lib/plugins/experience.js
redcarti: there might be user, not player.

src/lib/plugins/commands.js
redcarti: I think it might be ok

src/lib/plugins/moderation.js
redcarti: fixed

src/lib/plugins/moderation.js
redcarti: fixed

src/lib/plugins/moderation.js
redcarti: fixed

src/lib/plugins/particle.js
redcarti: only player

src/lib/plugins/commands.js
redcarti: excluding will reduce page length... so added player&console only

PR #405, opened 2020-06-25
1.13

Deudly:
Updating Flying-Squid to 1.13

- [x] Server starts
- [ ] diamond-square should be able to generate 1.13 maps and be compatible with < 1.13 versions [PR](https://github.com/PrismarineJS/diamond-square/pull/4)
- [ ] prismarine-chunk update: setBlockData() function for 1.13 [PR](https://github.com/PrismarineJS/prismarine-chunk/pull/90)
- [ ] Block IDs update (place)

Known issues:
- [ ] Only the person who loads the chunk can see the object data. Others (or on reconnect) will see the default block data
- [ ] Breaking a block in survival mode will result in a fatal error


src/lib/plugins/redstone.js
```diff
@@ -5,14 +5,18 @@ module.exports.server = function (serv, { version }) {
...
+  if (!serv.supportFeature('theFlattening')) {
+    unlitRedstoneTorchType = mcData.blocksByName.unlit_unlit_redstone_torch.id
+    poweredRepeaterType = mcData.blocksByName.powered_repeater.id
+    unpoweredRepeaterType = mcData.blocksByName.unpowered_repeater.id
+  }
   const powerLevel = (block, dir) => {
     if (block.type === redstoneWireType) return block.metadata
     if (block.type === redstoneTorchType) return 15
-    if (block.type === poweredRepeaterType) {
+    if (block.type === poweredRepeaterType || (block.type === repeaterType && block.metadata.powered == true)) {
```
Karang, line 19: `block.metadata` is a number, it doesn't have a powered property

src/lib/plugins/redstone.js
```diff
@@ -64,7 +68,7 @@ module.exports.server = function (serv, { version }) {
 
   const isDirectedRepeater = (block, dir, powered = false) => {
     if (block.type !== poweredRepeaterType &&
-        (block.type !== unpoweredRepeaterType || powered)) return false
+        (block.type !== unpoweredRepeaterType || block.type !== repeaterType || powered)) return false
```
Karang, line 71: I think a cleaner way would be to redefine the whole function, something like:
```js
const isDirectedRepeater = !serv.supportFeature('theFlattening') ?
(block, dir, powered = false) => {
  // current code
} :
(block, dir, powered = false) => {
 // 1.13+ code
}
```
Same for all the type helper functions:
* powerLevel
* powerLevelDir
* isWireDirectedIn
* etc..

src/lib/plugins/signs.js
```diff
@@ -1,7 +1,7 @@
 module.exports.server = (serv, { version }) => {
   const mcData = require('minecraft-data')(version)
 
-  const oakSignType = mcData.blocksByName.standing_sign.id
+  const oakSignType = mcData.blocksByName.sign.id
```
Karang, line 4: this change will break backward compatibility, a `if (serv.supportFeature('theFlattening'))` should probably be added

Deudly: Yeah, expect other changes that may break stuff. I didnt have the time to check the code and look for them

src/lib/plugins/redstone.js
Deudly: Now that we cant write it in a single line, doesnt it make more sense to use a more common syntax? The code you wrote is complex/weird, isnt it?

Karang: yeah I tried to be minimalist so you get the idea, in reality you would want to split in different variables

rom1504:
Prismarine-chunk is already compatible, what do you mean ?

Deudly:
> Prismarine-chunk is already compatible, what do you mean ?

@rom1504 Sorry, I meant updating some stuff that prismarine-chunk is missing, like https://github.com/PrismarineJS/prismarine-chunk/pull/90

src/lib/plugins/blocks.js
```diff
@@ -13,19 +13,19 @@ module.exports.player = function (player, serv) {
     else serv.updateBlock(player.world, position, serv.tickCount, serv.tickCount, true)
   }
 
-  player.sendBlock = (position, blockType, blockData) => // Call from player.setBlock unless you want "local" fake blocks
+  player.sendBlock = (position, blockType, blockData, blockStateId) => // Call from player.setBlock unless you want "local" fake blocks
```
Karang, line 16: you can remove blockType and blockData, it's redundant with blockStateId

src/lib/plugins/placeBlock.js
```diff
@@ -89,16 +89,15 @@ module.exports.player = function (player, serv, { version }) {
...
       if (playSound) {
         serv.playSound(sound, player.world, placedPosition.clone().add(new Vec3(0.5, 0.5, 0.5)), {
           pitch: 0.8
         })
       }
 
       if (player.gameMode === 0) { player.inventory.slots[36 + player.heldItemSlot]-- }
-
-      player.setBlock(position, id, damage)
+      player.setBlock(position, id, damage, heldItem.name)
```
Karang, line 100: You should retrieve the stateId from the name in placeBlock and only give the stateId to setBlock()

src/lib/plugins/redstone.js
Deudly: done

src/lib/plugins/redstone.js
Deudly: done

src/lib/plugins/blocks.js
Deudly: done

src/lib/plugins/blocks.js
```diff
@@ -13,19 +13,17 @@ module.exports.player = function (player, serv) {
...
+    }, ({ position }) => {
       player._client.write('block_change', {
         location: position,
-        type: blockType << 4 | blockData
+        type: blockStateId
       })
     })
 
-  player.setBlock = (position, blockType, blockData) => serv.setBlock(player.world, position, blockType, blockData)
+  player.setBlock = (position, stateId=0) => serv.setBlock(player.world, position, stateId)
```
rom1504, line 26: please remove that default value, it's error prone

PR #407, opened 2020-06-27
Tab completion, commands

redcarti:
I think that tab completion is done for now but other features are not.


redcarti:
oh what, why there's so many commits from my last PR...

redcarti:
ehh

PR #409, opened 2020-06-27
More work on 1.13

Karang:
Work in progress

PR #411, opened 2020-06-30
Save <player>.dat files (with player's inventory)

rrwr:
Continuation of #380


rom1504:
it seems test are failing

rrwr:
Looks like there is no settings.worldFolder when running tests

Karang:
> Looks like there is no settings.worldFolder when running tests

This will be fixed by https://github.com/PrismarineJS/flying-squid/pull/409

rom1504:
I reviewed this again.
Ultimately I think we'll want to move this in https://github.com/PrismarineJS/prismarine-provider-anvil but as a first step could you try to put the reading and the writing function in a file that just provides stateless function in lib/ folder, and then use it in plugin ?
That will make testing it much easier, and make the code cleaner

rrwr:
OK, but what do you mean by reading function? The code used in #380?

rom1504:
yes basically this part https://github.com/PrismarineJS/flying-squid/pull/380/files#diff-1dea21774fbc02bf339d7f711987fcbbR56

I think something that would take as input the player uuid and world folder and output an object that is just the keys to set in the player object would make a lot of sense
then you'd just loop over this object to set all these properties in the plugin code

src/lib/plugins/login.js
```diff
@@ -36,46 +31,31 @@ module.exports.server = function (serv, options) {
...
+    Object.keys(playerData.player).forEach(k => { player[k] = playerData.player[k] })
+
     serv.players.push(player)
     serv.uuidToPlayer[player.uuid] = player
     player.loadedChunks = {}
   }
 
   function updateInventory () {
-    playerSavedInventoryItems.forEach((item) => {
+    playerData.inventory.forEach((item) => {
```
rom1504, line 58: why not do this in addPlayer ? (that way you can make playerData local for addPlayer function)

rrwr: It seems like player.inventory.updateSlot doesn't work at the time when addPlayer is called

rrwr: player.inventory.updateSlot works only after sendLogin

rom1504: ok

rrwr:
It seems that now the tests should not fail (since #409 is already merged), can someone restart checks?

rom1504:
no, it's merged in 1.13 branch, not in master

rrwr:
Maybe it's better to rebase this pull request to 1.13 branch?

rom1504:
no, I just merged it to simplify

just rebase on master

rrwr:
ok

rom1504:
world should be saved only if region option is defined, you need to add an if

rrwr:
You mean that <player>.dat file should save only when something is in player.world.anvil.regions?

rom1504:
no, only when this option is not undefined https://github.com/PrismarineJS/flying-squid/blob/af52d3dca819aeada32838638cd4815bba590630/src/lib/plugins/world.js#L18

rom1504:
thanks !

Issue #413, opened 2020-07-02
Localization


redcarti:
Well, I've added ``locale.js`` plugin (PR will be soon), so I think commands and other things need to be localized.

Currently, addString looks like this:
```js
serv.locale.addString('lang', 'stringName', 'stringValue') // also stringName needs to be parsed like stringName/under or stringName:under, etc.
```

- [x] ``locale.js`` 
- [ ] localizations

mhsjlw:
Would it make sense to use an actual service for this where users can submit translations and localizations?

rom1504:
Mojang already provides all the translation, we can just use them.

What are you trying to do here ?

redcarti:
Mojang provides translations for minecraft commands but contributors can make own plugins with own localization

redcarti:
But flying-squid doesn't have any localization now. So it kinda make sense..?

redcarti:
All objects (needs to find langs): `.minecraft/assets/indexes`,
Searching in `.minecraft/assets/objects`

lastuniverse:
Sorry. Can this module help solve this problem?
[https://www.npmjs.com/package/msg-localize](https://www.npmjs.com/package/msg-localize)

Unfortunately, I got into the essence of the problem very superficially.
Perhaps I could remake it so that it solves the task, but for this I need to explain the essence of the task in a little more detail)

lastuniverse:
## task
```javascript
serv.locales.setStringLang(lang, path, value)
serv.locales.setString(path, values)
serv.locales.getString(lang, path) || serv.localeString(lang, path)
player.localeString(path)
```


## Your еxamples

#### Set string in one lang

```javascript
serv.locales.setStringLang('en_us', 'example.path', 'Hello!')
```

####  Set string to multiple langs
```javascript
serv.locales.setString('example.path.multiple', {
  en_US: 'Multiple languages',
  ru_RU: 'Несколько языков'
})
```

#### Get localized string
```javascript
serv.info(serv.localeString('en_US', 'example.path.multiple')) // return 'Multiple languages'
```

## Examples with [msg-localize](https://www.npmjs.com/package/msg-localize)

#### Set string in one lang 
in file `path/to/lang/folder/en_us.json`
```javascript
{
    "example": {
        "path": "Hello!"
    }
}
```
#### Set string to multiple langs
in file `path/to/lang/folder/en_us.json`
```javascript
{
    "example": {
        "path": "Hello!"
    }
}
```

[...]

PR #414, opened 2020-07-02
Localization

redcarti:
PR of #413 

* [x] ``serv.locales.setStringLang(lang, path, value)``
* [x] ``serv.locales.setString(path, values)``
* [x] ``serv.locales.getString(lang, path)`` || ``serv.localeString(lang, path)``
* [x] ``player.localeString(path)``
* [x] local localization
  1. [x] localizations
  2. [x] add to ``serv.locales.langs``
* [x] docs
* [x] tests 

### Examples
Set string in one lang
```js
serv.locales.setStringLang('en_us', 'example.path', 'Hello!')
```

Set string to multiple langs
```js
serv.locales.setString('example.path.multiple', {
  en_US: 'Multiple languages',
  ru_RU: 'Несколько языков'
})
```

Get localized string
```js
serv.info(serv.localeString('en_US', 'example.path.multiple')) // return 'Multiple languages'
```

### Info
``serv.locales.setString`` replaces values.


package.json
```diff
@@ -31,6 +31,7 @@
     "emit-then": "^2.0.0",
     "event-promise": "^0.0.1",
     "flatmap": "^0.0.3",
+    "lodash": "^4.17.15",
```
rom1504, line 34: I don't think we really need lodash for this

rom1504:
can you rebase on master

redcarti:
will rebase it later

package.json
redcarti: I used lodash for get & set strings

redcarti:
> where does the locale information come from ?

player locale? `settings` packet. or sth other?

package.json
rom1504: yes, you can do it without using that lib, it's pretty big for this basic use case

rom1504:
ah yeah right ok

package.json
redcarti: hmm, well, will rewrite later

rom1504:
the locale files should be in minecraft-data and can you PR any adaptation you made to the extractor in minecraft jar extractor ?

redcarti:
> the locale files should be in minecraft-data and can you PR any adaptation you made to the extractor in minecraft jar extractor ?

I didn't use jar extractor, en-us lang stored only.

rom1504:
> I didn't use jar extractor, en-us lang stored only.

ok could you adapt it and do it ?
doing this manually is quite annoying to maintain

redcarti:
> ok could you adapt it and do it ?
> doing this manually is quite annoying to maintain

well I can create PR to mcdata and store langs there but it needs to be finded

rom1504:
Do you think it would be very hard to adapt https://github.com/PrismarineJS/minecraft-jar-extractor/blob/master/lang.js to work for more languages ?

redcarti:
I mean jar hasn't any langs, except en_us. So, as I said in #413, they stored locally in assets folder

redcarti:
as I know.. but I checked jars and didn't found any langs

rom1504:
asset folder can be downloaded using https://github.com/PrismarineJS/node-minecraft-wrap#launcherdownloadgetassetassetfileversion
see this example https://github.com/PrismarineJS/node-minecraft-wrap/blob/master/examples/exampleLauncherDownload.js

redcarti:
oh well, okay. will try.

What about `localeBroadcast()`? I've tested it but the code is not that good...

rom1504:
the languages stored should be in minecraft-data, not here https://github.com/PrismarineJS/minecraft-jar-extractor/pull/23#issuecomment-658254528

rom1504:
@redcarti tests are failing ? any plan to finish this ?

redcarti:
I wanted to log that langs are loaded and etc. but log.js plugin loaded after locale.js, so it kinda broked

redcarti:
I think that log.js can be moved in index.js, where server is starting

rom1504:
avoid to do things that plugin loading time and instead do it after an event is triggered for example

redcarti:
okay, well

redcarti:
I forgot about it, ehh.. So, it fails cuz I'm logging `Downloading langs...` & `Loading langs...` with `Loaded ${count} langs` and it says "Cannot log" or something like that

redcarti:
Well, idk how to solve it...

PR #419, opened 2020-07-04
Fix portal command

Karang:


rom1504:
I don't think this is enough.
Try it, do /portal something then take a flint and steel and activate it inside the portal.
It does incorrect things currently
(when fixed it'll allow us to test world changing)

Issue #416, opened 2020-07-04
Fix a few remaining 1.13 issues


rom1504:
Noticed issues :

* [ ] no clear error for incorrect saved world version (not blocking)
* [x] block dug drop is broken (go to /gamemode 0 to test)
* [x] dropping an item from hotbar (press q) crash the server
* [x] portal is using incorrect blocks (can be tested with the /portal command)

Karang:
The broken drop is not specific to 1.13, and you can test it in creative by pressing the drop key (q)

Dropping should be handled in https://github.com/PrismarineJS/flying-squid/blob/master/src/lib/plugins/digging.js#L8 (status 4)
All versions >= 1.8 should be the same: https://wiki.vg/index.php?title=Protocol&oldid=7368#Player_Digging

rom1504:
https://github.com/PrismarineJS/flying-squid/blob/303ad467c987487b2f2b46d89e89dc5cf9bb01e3/src/lib/plugins/digging.js#L134

it is handled, but crashing the server in 1.13

rom1504:
ah I see, you're saying the server is sending us that status == 4 in all cases ?
then it's handled incorrectly indeed
it used to work (even if not vanilla then) for 1.8 and 1.12

Karang:
Mmm wait, are we talking about the same thing ? I'm talking about when you drop an item from your inventory. Are you talking about the block that drop when you break a block ?

rom1504:
ah, yes I was talking about drop when breaking a block, not inventory

rom1504:
Done, moving last point to https://github.com/PrismarineJS/flying-squid/issues/421

PR #420, opened 2020-07-05
Fix item drops and a few bugs along the way

Karang:
Fix 2 issues of #416 +
* Placed blocks were not correctly removed from inventory in survival
* Using /gamemode 0 would set the string "0" in player.gameMode, making the player.gameMode===0 test fail
* Picked items now go in the hotbar when possible

### DIFF
--- a/src/lib/plugins/digging.js
+++ b/src/lib/plugins/digging.js
@@ -1,43 +1,61 @@
 const Vec3 = require('vec3').Vec3
 
-module.exports.player = function (player, serv) {
+module.exports.player = function (player, serv, { version }) {
   function cancelDig ({ position, block }) {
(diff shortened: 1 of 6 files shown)


src/lib/plugins/digging.js
```diff
@@ -1,43 +1,61 @@
...
         block = facedBlock
         pos = facedPos
-      } else block = await player.world.getBlock(pos)
+      } else {
+        block = await player.world.getBlock(pos)
+      }
 
       currentlyDugBlock = block
       if (currentlyDugBlock.type === 0) return
-      if (status === 0 && player.gameMode !== 1) {
```
rom1504, line 24: Great, one less behavior

PR #430, opened 2020-08-05
Initial support of 1.14, 1.15 and 1.16

### LINKED ISSUE #422: Support more recent versions : 1.14, 1.15, 1.16
Should be much easier than 1.13 since we're now paste the flatening
Corresponding mineflayer PRs will help

IdanHo:
1.14, 1.15 and 1.16 support is mostly done. closes #422 
Cannot be merged until the following 2 packages are updated:
- prismarine-provider-anvil: updated to support 1.14, 1.15 and 1.16 (as the server cannot save its state to disk currently, so on restart all changes are lost). See PrismarineJS/prismarine-provider-anvil#25
- prismarine-chunk: updated to produce the 'MOTION_BLOCKING' heightmaps (weirdly packed long arrays), but this is less important, as the fake heightmaps i currently send work fine for everything but a couple of esoteric features, like dogs getting wet in the rain.

tools/dumpCodec is a script that uses minecraft-wrap to dump the latest dimensionCodec, as its constant


IdanHo:
Tests are failing as there's only 3 tests but 5 testing vms i think, so not related to this pull

rom1504:
Tests failing here are definitely related to this PR, have a look again

rom1504:
You need to add one number in https://github.com/PrismarineJS/flying-squid/blob/master/.github/workflows/ci.yml#L17 and the 6 at the bottom

IdanHo:
> You need to add one number in https://github.com/PrismarineJS/flying-squid/blob/master/.github/workflows/ci.yml#L17 and the 6 at the bottom

tests are failing due to "Your test suite must contain at least one test.", how would adding another circle node fix that?

IdanHo:
well increasing the node count did indeed fix it, but looking at parallel.js's source i cant tell why, from what i understand it should just have allocated the 1.14 tests to one of the nodes so it would have done 2 tests instead of 1. anyhow, all of the tests are passing besides some 1.14 ones due to prismarine-provider-anvil not having 1.14 support. if you noop out the writing to disk of prismarine-provider-anvil it passes.

IdanHo:
All of the tests pass locally when you disable the usage of provider-anvil and you can connect and play with the vanilla minecraft normally, so all that is left now is to update prismarine-provider-anvil.

rom1504:
Could you check the following features still work in 1.16:
* commands (test a few of them)
* position sync (connect 2 clients and make sure you can see each other)
* Digging
* Placing blocks

IdanHo:
tested all of those, and all of them indeed work, but i found a couple of small bugs along the way, committing the fixes now, but this should not merged until the anvil merge is done.

rom1504:
I released anvil 2.3.0 you can update it here

PR #431, opened 2020-08-08
Swap attribute names

mdashlw:


rom1504:
@IdanHo does that look right to you ?

IdanHo:
Yep, not sure how i managed to flip that.

PR #433, opened 2020-08-08
Added functionality to switch items from offhand and current hand slot when pressing F

tumenor:


tumenor:
Gotta change some dirty code added for testing.

tumenor:
Should be correct, pardon my issues. I've never used a pull request before ^^

PR #437, opened 2020-08-09
Fix crash in /gamemode command

lleyton:
Currently, if a player issues /gamemode with an invalid game mode which is not a number (Ex: "/gamemode survival") the server crashes. This PR modifies a check to validate that the inputted game mode is a number from 0-3.

### DIFF
--- a/src/lib/plugins/players.js
+++ b/src/lib/plugins/players.js
@@ -24,8 +24,8 @@ module.exports.server = function (serv, { version }) {
       if (paramsSplit[0] === '') {
         return false
       }
-      if (!paramsSplit[0].match(/^([0-3])$/) && paramsSplit[0].match(/^([0-9]+)$/)) {
-        throw new UserError(`The number you have entered (${paramsSplit[0]}) is too big, it must be at most 3`)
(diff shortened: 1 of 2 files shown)


rom1504:
Don't you think it would make more sense to handle /gamemode survival ?

lleyton:
Yeah, that makes sense.

src/lib/plugins/players.js
```diff
@@ -17,45 +17,46 @@ module.exports.server = function (serv, { version }) {
     base: 'gamemode',
     aliases: ['gm'],
     info: 'to change game mode',
-    usage: '/gamemode <0-3> [player]',
+    usage: '/gamemode <mode> [player]',
```
rom1504, line 20: seems you now removed support completely for 0-3.
Vanilla handles both 0-3 and survival right?
maybe we could do the same?

lleyton: Since 1.13, vanilla stopped handling numeric game mode IDs.

SiebeDW: what about supporting them both?

PR #438, opened 2020-08-09
Use bedrock instead of polished andesite

GroobleDierne:
I know it's hard to make a perfect vanilla generation, but maybe using bedrock instead of polished andesite will be nice :)


rom1504:
We recently moved out of hard coding all the IDs. It seems there are still a few lying around.
Thanks for fixing this one

PR #439, opened 2020-08-09
Blocks id

GroobleDierne:
Use minecraft-data to get blocks id

### DIFF
--- a/src/lib/worldGenerations/grass_field.js
+++ b/src/lib/worldGenerations/grass_field.js
@@ -3,14 +3,15 @@ const Vec3 = require('vec3').Vec3
 function generation ({ version }) {
   const Chunk = require('prismarine-chunk')(version)
   const mcData = require('minecraft-data')(version)
+  const theFlattening = require('../../../app').supportFeature('theFlattening')
 
   function generateSimpleChunk () {
     const chunk = new Chunk()
 
     for (let x = 0; x < 16; x++) {
       for (let z = 0; z < 16; z++) {
-        chunk.setBlockType(new Vec3(x, 50, z), mcData.blocksByName.grass_block.id)
-        chunk.setBlockData(new Vec3(x, 50, z), 1)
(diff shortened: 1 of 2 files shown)


src/lib/worldGenerations/grass_field.js
```diff
@@ -2,13 +2,15 @@ const Vec3 = require('vec3').Vec3
...
   const Chunk = require('prismarine-chunk')(version)
+  const mcData = require('minecraft-data')(version)
 
   function generateSimpleChunk () {
     const chunk = new Chunk()
 
     for (let x = 0; x < 16; x++) {
       for (let z = 0; z < 16; z++) {
-        chunk.setBlockType(new Vec3(x, 50, z), 2)
+        chunk.setBlockType(new Vec3(x, 50, z), mcData.blocksByName.grass_block.id)
```
Karang, line 12: Careful, in old version grass_block is grass

GroobleDierne: Do you know which version ?

Karang: Bellow 1.13

GroobleDierne: Do you recommend to use serv.supportFeature() or a normal condition ?

rom1504: serv.supportFeature would be good, I believe we already have something similar in the code

GroobleDierne:
it's good for me.

src/lib/worldGenerations/grass_field.js
```diff
@@ -2,13 +2,16 @@ const Vec3 = require('vec3').Vec3
 
 function generation ({ version }) {
   const Chunk = require('prismarine-chunk')(version)
+  const mcData = require('minecraft-data')(version)
+  const theFlattening = require('../../../app').supportFeature('theFlattening')
```
rom1504, line 6: this important would start the server, that's not good.
Can you directly important the right thing?

rom1504 (approved):

PR #440, opened 2020-08-22
Remove deprecated/heavy dependencies (request, request-promise) in favor of needle

WasabiThumb:
Request and request-promise are both required for use in a single file. This is wasteful (and ill-advised, as the packages are deprecated), so I've decided to replace it with needle. You could probably also use the native http library, but I would argue importing needle is worth the trouble of adding error handling and json detection.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "minecraft-protocol": "^1.15.0",
     "mkdirp": "^0.5.1",
     "moment": "^2.10.6",
+    "needle": "^2.5.0",
     "node-gzip": "^1.1.2",
     "prismarine-chunk": "^1.20.1",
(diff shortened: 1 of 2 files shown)


rom1504 (approved):

rom1504:
thanks

rom1504:
if you like this kind of fixing, you may like https://github.com/PrismarineJS/prismarine-provider-anvil/issues/28

Karang:
> if you like this kind of fixing, you may like [PrismarineJS/prismarine-provider-anvil#28](https://github.com/PrismarineJS/prismarine-provider-anvil/issues/28)

Ads got to an other level :D

PR #451, opened 2020-10-20
Block entities

mat-1:
Allows block entities to work (currently only implemented signs, but it shouldn't be too much work to get other ones working)

Related: PrismarineJS/prismarine-chunk#102 & PrismarineJS/prismarine-provider-anvil#31


mat-1:
Block entities now works (only for signs because who cares about the *other* block entities)
PrismarineJS/prismarine-chunk#102 and PrismarineJS/prismarine-provider-anvil#31 must be merged for it to actually work

Issue #461, opened 2020-12-21
1.12.2 - Chunks are not sending


TheAlan404:
No error in logs, tried both superflat and diamond-square, im not recieving any chunks...

Did a test with proxy.js --dump-all. No chunk related packets whatsoever...

rom1504:
fixed now

TheAlan404:
I am still not getting any chunks... update didnt work

PR #462, opened 2020-12-21
Fix header.js and superflat generator

TheAlan404:
Header.js - Added defaults so 'player-header-text' is now all optional
Superflat - Fixed require() path so it doesnt crash

Tested it locally, both works. Hope this is good enough

### DIFF
--- a/src/lib/plugins/header.js
+++ b/src/lib/plugins/header.js
@@ -6,6 +6,6 @@ module.exports.player = function (player, serv, settings) {
     })
 
   if (settings['player-list-text']) {
-    player.playerlistUpdateText(settings['player-list-text'].header || { text: '' }, settings['player-list-text'].footer || { text: ''})
+    player.playerlistUpdateText(settings['player-list-text'].header || { text: '' }, settings['player-list-text'].footer || { text: '' })
   }
 }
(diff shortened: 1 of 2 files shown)


src/lib/plugins/header.js
```diff
@@ -5,5 +5,7 @@ module.exports.player = function (player, serv, settings) {
       footer: JSON.stringify(footer)
     })
 
-  player.playerlistUpdateText(settings['player-list-text'].header, settings['player-list-text'].footer)
+  if (settings['player-list-text']) {
```
rom1504, line 8: Can you make it use good defaults instead of not calling the method at all ?

TheAlan404: Ok sure

dada513: vanilla does not always use a custom player list so in my opinion its better that way

dada513 (approved):

src/lib/plugins/header.js
rom1504: Ah yeah then fine let's keep it that way

TheAlan404: Okay, reverted it.

rom1504:
Ci is failing

dada513 (approved):
PR good though you should take a look at the CI since its failing

TheAlan404:
I'm pretty sure the fail is related to issue #461 though...

rom1504 (approved):

PR #475, opened 2021-02-12
Adds /clear

u9g:
It's not done yet because if you give `/clear [username] [block] [number]` the number part on L49 - L 120 get's confused, everything else should work

### DIFF
--- a/src/lib/plugins/clear.js
+++ b/src/lib/plugins/clear.js
@@ -91,9 +91,9 @@ module.exports.server = (serv, { version }) => {
             }
           }
         }
-        // check inventory
         if (currCount !== count) {
-          for (let i = player.inventory.inventoryStart; i < player.inventory.inventoryEnd - 1; i++) {
+          // start before hotbar
+          for (let i = player.inventory.inventoryStart; i < player.inventory.inventoryEnd - 10; i++) {
             const currSlot = player.inventory.slots[i]
             const blocksNeeded = count - currCount
             if (currSlot?.type === BLOCK_ID) {
@@ -138,8 +138,6 @@ module.exports.server = (serv, { version }) => {
         }
 
         return clearedCount
-        // implement picking which stacks to take from
-        // serv._writeArray(packetName, packetFields, playerArray)
       } else {
         const BLOCK_ID = mcData.blocksByName[blockType].id
         const blocks = player.inventory.slots
--- a/src/lib/plugins/clear.js
+++ b/src/lib/plugins/clear.js
@@ -11,8 +11,12 @@ module.exports.server = (serv, { version }) => {
     base: 'clear',
     info: "Clear a player's inventory.",
     usage: '/clear <player> <item> <maxCount>',
-    parse: /(?:(\w+)(?: (\w+)(?: (\d+))?)?)?/, // ex: /clear USERNAME_ stone 1
+    parse: /^(\w+)?(?: (\w+))?(?: (\d+))?$/, // ex: /clear USERNAME_ stone 1
     action (params, ctx) {
+      if (typeof params === 'string') { // the parsing failed so it just returned initial input
+        return `${this.usage}: ${this.info}`
+      }
+
       Item = require('prismarine-item')(version)
       mcData = require('minecraft-data')(version)
       // if the ctx.player doesn't exist, it's console executing the command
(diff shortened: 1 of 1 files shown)


src/lib/plugins/clear.js
```diff
@@ -0,0 +1,172 @@
...
+        const BLOCK_ID = mcData.blocksByName[blockType].id
+        let currCount = 0
+        const partial = {
+          used: false,
+          item: null,
+          slot: 0
+        }
+        const slots = []
+        // check hotbar first from right to left
+        for (let i = player.inventory.inventoryEnd - 1; i > player.inventory.inventoryEnd - 10; i--) {
```
Karang, line 69: You can use `findItemRange` or `findItemRangeName` https://github.com/PrismarineJS/prismarine-windows/blob/master/lib/Window.js#L158 for this.

Use `player.inventory.hotbarStart` instead of -10.

Why iterating in reverse ? Is that how clear work ?

src/lib/plugins/clear.js
```diff
@@ -0,0 +1,172 @@
...
+                clearedCount += currSlot.count
+                slots.push(i)
+                break // we have enough items
+              } else { // too little items to finish counter
+                currCount += currSlot.count
+                clearedCount += currSlot.count
+                slots.push(i)
+              }
+            }
+          }
```
Karang, line 120: Big chunk of duplicated code, put it in a function.

src/lib/plugins/clear.js
```diff
@@ -0,0 +1,172 @@
...
+      for (const { slot } of blocks) {
+        serv._writeArray(PACKET_NAME, {
+          windowId: 0,
+          slot: slot,
+          item: Item.toNotch(AIR_ITEM)
+        }, playerArray)
+        player.inventory.updateSlot(slot, AIR_ITEM)
+      }
+      return totalBlocksCleared
+    }
```
Karang, line 170: Very similar to the code above, put in a function.

src/lib/plugins/clear.js
```diff
@@ -0,0 +1,172 @@
...
+    info: "Clear a player's inventory.",
+    usage: '/clear <player> <item> <maxCount>',
+    parse: /^(\w+)?(?: (\w+))?(?: (\d+))?$/, // ex: /clear USERNAME_ stone 1
+    action (params, ctx) {
+      if (typeof params === 'string') { // the parsing failed so it just returned initial input
+        return `${this.usage}: ${this.info}`
+      }
+
+      Item = require('prismarine-item')(version)
+      mcData = require('minecraft-data')(version)
```
Karang, line 21: Item and mcData should be loaded outside the command actions + there is no need for them to be in global scope.

src/lib/plugins/clear.js
```diff
@@ -0,0 +1,172 @@
...
+        }
+      } else {
+        const blocksCleared = clearInventory(executingPlayer, undefined, undefined)
+        return `Removed ${blocksCleared} from player ${executingPlayer.username}`
+      }
+    }
+  })
+
+  function clearInventory (player, blockType, count) {
+    const AIR_ITEM = null
```
Karang, line 54: No need to define this constant, just use null where relevant. (Its not an AIR_ITEM, its an empty slot, aka null)

Karang (changes requested):

src/lib/plugins/clear.js
u9g: Yeah, it goes in reverse on vanilla

u9g: Also, that method doesn't help me with anything

src/lib/plugins/clear.js
u9g: fixed

src/lib/plugins/clear.js
u9g: fixed

src/lib/plugins/clear.js
```diff
@@ -0,0 +1,172 @@
...
+    const PACKET_NAME = 'set_slot' // https://wiki.vg/Protocol#Set_Slot & https://minecraft-data.prismarine.js.org/?d=protocol#toClient_set_slot
```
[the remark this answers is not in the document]
u9g: fixed

u9g:
closed for https://github.com/PrismarineJS/flying-squid/pull/478

PR #476, opened 2021-02-12
Fixed useItem

KaffinPX:
useItem.js was broken i fixed it for multiple versions.

### DIFF
--- a/src/lib/plugins/useItem.js
+++ b/src/lib/plugins/useItem.js
@@ -2,7 +2,7 @@ module.exports.server = (serv, { version }) => {
   const mcData = require('minecraft-data')(version)
   const mobs = mcData.mobs
 
-  function getEntID (entName) { // 1.9x, 1.10x and 1.11x
+  function getEntID(entName) { // 1.9x, 1.10x and 1.11x
     let foundID = ''
 
     Object.keys(mobs).forEach(mobID => {
@@ -14,8 +14,8 @@ module.exports.server = (serv, { version }) => {
 
     return foundID
   }
-  
-  function getEntIDmc (entName) { // 1.12x
+
+  function getEntIDmc(entName) { // 1.12x
     let foundID = ''
 
     Object.keys(mobs).forEach(mobID => {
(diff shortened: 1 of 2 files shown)


src/lib/plugins/useItem.js
```diff
@@ -27,10 +40,24 @@ module.exports.server = (serv, { version }) => {
...
+        })
+      } else {
+        if (version === '1.8.8') {
+          serv.onItemPlace('spawn_egg', ({ item, player, placedPosition }) => {
+            serv.spawnMob(item.metadata, player.world, placedPosition)
+            return { id: -1, data: 0 }
+          })
+        } else {
+          serv.onItemPlace('spawn_egg', ({ item, player, placedPosition }) => {
+            serv.spawnMob(getEntID(item.nbt.value.EntityTag.value.id.value), player.world, placedPosition)
```
Karang, line 56: You could have only 1 getEntID function if you call it like:
```js
serv.spawnMob(getEntID('minecraft:' + item.nbt.value.EntityTag.value.id.value), player.world, placedPosition)

```

src/lib/plugins/useItem.js
```diff
@@ -27,10 +40,24 @@ module.exports.server = (serv, { version }) => {
...
-        serv.spawnMob(getEntID(heldItem.nbt.value.EntityTag.value.id.value), player.world, placedPosition)
-        return { id: -1, data: 0 }
-      })
+      if (version === '1.12.2') {
+        serv.onItemPlace('spawn_egg', ({ item, player, placedPosition }) => {
+          serv.spawnMob(getEntIDmc(item.nbt.value.EntityTag.value.id.value), player.world, placedPosition)
+          return { id: -1, data: 0 }
+        })
+      } else {
+        if (version === '1.8.8') {
```
Karang, line 49: To check the version you need to use the supportFeature system (it must also work for other 1.8 versions (snapshots etc..)

Basically, you create an entry in https://github.com/PrismarineJS/flying-squid/blob/master/src/lib/features.json#L63
And check with
```js
if (serv.supportFeature('entityMCPrefixed')) {
```

src/lib/plugins/useItem.js
KaffinPX: Optimizing code wait a minute

src/lib/plugins/useItem.js
KaffinPX: Applying...

rom1504:
Can you add a test so it doesn't get broken again in the future ?

rom1504 (approved):

PR #478, opened 2021-02-13
adds /clear and server messages and p-chat and a better way to parse selector strings

u9g:


u9g:
This won't pass tests until this is merged https://github.com/PrismarineJS/prismarine-windows/pull/21

Karang:
> This won't pass tests until this is merged [PrismarineJS/prismarine-windows#21](https://github.com/PrismarineJS/prismarine-windows/pull/21)

You can update package.json to point to your github branch `"prismarine-windows": "u9g/prismarine-windows#patch-2"`

src/lib/plugins/clear.js
```diff
@@ -0,0 +1,58 @@
...
+        return player.inventory.clear(BLOCK_ID, count)
+      } else {
+        const BLOCK_ID = mcData.blocksByName[blockType].id
+        return player.inventory.clear(BLOCK_ID)
+      }
+    } else {
+      return player.inventory.clear()
+    }
+  }
+}
```
Karang, line 58: The whole function is equivalent to
```js
function clearInventory (player, blockType, count) {
  const id = blockType ? mcData.blocksByName[blockType].id : null
  return player.inventory.clear(id, count)
}
```

src/lib/plugins/inventory.js
```diff
@@ -8,6 +8,14 @@ module.exports.player = function (player, serv, { version }) {
   player.heldItem = new Item(256, 1)
   player.inventory = windows.createWindow(1, 'minecraft:inventory', 'inv', 36)
 
+  player.inventory.on('updateSlot', (slot, newItem) => {
+    player._client.write('set_slot', {
+      windowId: 0, // inventory window id
```
Karang, line 13: Its correct, but what about making it more explicit as:
```js
windowId: player.inventory.id
```
You won't need the commentary

src/lib/plugins/inventory.js
```diff
@@ -8,6 +8,14 @@ module.exports.player = function (player, serv, { version }) {
   player.heldItem = new Item(256, 1)
   player.inventory = windows.createWindow(1, 'minecraft:inventory', 'inv', 36)
 
+  player.inventory.on('updateSlot', (slot, newItem) => {
```
Karang, line 11: Incorrect, the arguments are: `(slot, oldItem, newItem)`

src/lib/plugins/clear.js
```diff
@@ -0,0 +1,58 @@
...
+            } else { // everything valid not defined
+              const blocksCleared = clearInventory(serv.getPlayer(wantedUsername), params[2], parseInt(params[3]))
+              return `Removed ${blocksCleared} items from player ${wantedUsername}`
+            }
+          }
+        } else {
+          if (executingPlayer) { // player executing on themselves
+            const blocksCleared = clearInventory(executingPlayer, undefined, undefined)
+            return `Removed ${blocksCleared} items from player ${executingPlayer.username}`
+          } else { // console executing with no player arg
```
Karang, line 33: Don't use else if you have a return in the if.

Karang (changes requested):

src/lib/plugins/clear.js
Karang: Given the function is so short, I'm not sure its needed.

u9g:
updated clear.js to query itemsbyname instead of blocksbyname so you can clear items too

u9g:
example admin images: 
image
image

u9g:
admin messages rely on https://github.com/PrismarineJS/prismarine-chat/pull/10 so I added that to deps

src/lib/plugins/clear.js
```diff
@@ -0,0 +1,186 @@
...
+    usage: '/clear <player> <item> <maxCount>',
+    parse: /^((?:@|\w)+)?(?: (\w+))?(?: (\d+))?$/, // ex: /clear USERNAME_ stone 1
+    action (params, ctx) {
+      const playersString = (players) => players.length > 1 ? `${players.length} players` : `player ${players[0].username}`
+      const isConsole = Object.entries(ctx).length === 0
+      const isOp = isConsole ? true : ctx.player.op
+      const notUdf = (x) => x !== undefined
+      console.log()
+      if (typeof params === 'string') return `${this.usage}: ${this.info}` // parsing failed
+      if (isConsole && !notUdf(params[1])) return 'A player is required to run this command here' // console
```
extremeheat, line 16: You can destructure the array to make it easier to follow, like `const [ a, b, c ] = [1, 2]`

u9g:
TODO: 
1. figure out selector string parsing for console
2. remove default sender and position in sendChatMessage and declare them in the call to the function

src/lib/plugins/clear.js
```diff
@@ -0,0 +1,186 @@
...
+      if (players.some(player => player == null)) return "Player given doesn't exist"
+      const id = params[2] ? mcData.itemsByName[params[2]].id : null
+      // also clear held item if players are in survival
+      const blocksCleared = players.reduce((accumulator, player) => {
+        const res = player.inventory.clear(id, params[3] ? parseInt(params[3]) : undefined)
+        return accumulator + res
+      }, 0)
+      if (blocksCleared === 0) return `No items were found on ${playersString(players)}`
+      let serverMessage
+      if (!isConsole && players.length > 1) serverMessage = execByUserToUsers(ctx.player.username, ctx.player.uuid, blocksCleared, players.length)
```
extremeheat, line 34: The logic here is hard to follow, break up the if statements with braces, and add some line spacing to separate the logic. I think it the logic here could be refactored and simplified, but that's up to you.

src/lib/plugins/commands.js
```diff
@@ -335,6 +335,30 @@ module.exports.server = function (serv, { version }) {
...
+  /**
+   * @description send to a user a prismarine-chat message
+   * @param {string} aboutUser user the message is about
+   * @param {boolean} isConsole executed by console (alias for showToUser to be true)
+   * @param {ChatMessage} message message sent to user
+   * @param {boolean} showToUser show the message to @var aboutUser
+   * @param {string} sender usually is the preset
+   * @param {int} position usually is the preset
+   */
+  serv.sendChatMessage = (aboutUser, isConsole, message, showToUser, sender = '00000000-0000-0000-0000-000000000000', position = 1) => {
```
extremeheat, line 352: You can import the uuid library and do `uuid.v3()` to generate a random UUID here, otherwise you should define the default constant at the top of the file. Also, why the need for two parameters that do the same thing? The logic could be simplified here

src/lib/plugins/commands.js
```diff
@@ -335,6 +335,30 @@ module.exports.server = function (serv, { version }) {
...
   }
 
+  serv.parseSelectorString = (str, pos, world, allowUser) => {
+    if (/^@/.test(str) && !(/^@([arpe])(?:\[([^\]]+)\])?$/.test(str))) throw new UserError('Unknown selector type')
+    if (!(/^@/.test(str))) { return { str: serv.getPlayer(str) } }
+    return serv.selectorString(str, pos, world, allowUser)
+  }
+  /**
+   * @description send to a user a prismarine-chat message
+   * @param {string} aboutUser user the message is about
```
extremeheat, line 345: This should be the player uuid, not username

extremeheat (changes requested):
Overall ok. I think you did over engineer some things, and there are some readability issues, left some comments

src/lib/plugins/commands.js
u9g: No, it's correct, this was done on purpose.

src/lib/plugins/commands.js
u9g: it's not random

u9g:
I'm just going to leave it like this, but eventually someone will have to remake the selectorstring function to support console using it, and at that point, this command should be changed to support that.

src/lib/plugins/clear.js
```diff
@@ -0,0 +1,204 @@
+module.exports.server = (serv, { version }) => {
+  const mcData = require('minecraft-data')(version)
+  const ChatMessage = require('prismarine-chat')(version)
+  const notUdf = x => x !== undefined
```
Karang, line 4: Why define a function for this ? Just inline it.

src/lib/plugins/clear.js
```diff
@@ -0,0 +1,204 @@
...
+      // player(s) to execute on
+      let players
+      if (isConsole) {
+        players = [serv.getPlayer(playerSelector)]
+      } else if (notUdf(playerSelector)) {
+        players = Object.entries(serv.parseSelectorString(playerSelector, ctx.player.position, ctx.player.world)).map(p => p[1])
+      } else { // no input given (so default to executor)
+        players = [ctx.player]
+      }
+      if (players.some(player => player == null)) return "Player given doesn't exist"
```
Karang, line 30: All this player selection logic looks like it could be put in its own function to be reused for other commands.

src/lib/plugins/clear.js
```diff
@@ -0,0 +1,204 @@
...
+                  }
+                }
+              },
+              text: toUser
+            }
+          ]
+        }
+      ]
+    })
+  }
```
Karang, line 146: These chat messages should be treated as data and go in a separate json file.

src/lib/plugins/login.js
```diff
@@ -77,6 +77,10 @@ module.exports.player = async function (player, serv, settings) {
       const slot = convertInventorySlotId.fromNBT(item.Slot.value)
       player.inventory.updateSlot(slot, newItem)
     })
+    for (let ix = player.inventory.inventoryStart; ix < player.inventory.inventoryEnd; ix++) {
+      const slot = player.inventory.slots[ix]
+      if (!slot) player.inventory.updateSlot(ix, null)
+    }
```
Karang, line 83: This isn't needed anymore, with your pr in prismarine-windows

Karang (changes requested):

src/lib/plugins/clear.js
u9g: done

src/lib/plugins/clear.js
u9g: done

src/lib/plugins/inventory.js
```diff
@@ -6,7 +6,7 @@ module.exports.player = function (player, serv, { version }) {
 
   player.heldItemSlot = 0
   player.heldItem = new Item(256, 1)
-  player.inventory = windows.createWindow(1, 'minecraft:inventory', 'inv', 36)
+  player.inventory = windows.createWindow(0, 'minecraft:inventory', 'inv', 36)
```
extremeheat, line 9: can you explain this change?

extremeheat (changes requested):

src/lib/plugins/commands.js
```diff
@@ -335,6 +335,63 @@ module.exports.server = function (serv, { version }) {
...
+        )
+        .map(p => p[1])
+    } else { // no input given (so default to executor)
+      players = [ctx.player]
+    }
+    return players
+  }
+
+  serv.parseSelectorString = (str, pos, world, allowUser) => {
+    if (/^@/.test(str) && !(/^@([arpe])(?:\[([^\]]+)\])?$/.test(str))) throw new UserError('Unknown selector type')
```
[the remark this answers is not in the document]
u9g: Eventually, yes we could do that, but that feels like a whole different pr then this, this regex is literally taken from an earlier point in the same file. Yes it's messy, but if we merge this. we will eventually have to move all things to brigadier anyway.

src/lib/plugins/inventory.js
u9g: this is supposed to be 0 https://github.com/PrismarineJS/prismarine-windows/blob/7798a6858667c9cdf5d11d11a7101641c38b3cfb/index.js#L11

rom1504:
should we close this?

andriycraft:
> should we close this?

no

andriycraft (commented):
This pull request is too old
the last commit was on 7 Apr 2021

rom1504:
@u9g what do you think ?

andriycraft:
@u9g will u finish this pr?

rom1504:
closed with label "was good"

Maybe someone will pick it up someday

PR #481, opened 2021-02-15
on login set unused slots to null

u9g:


extremeheat:
Why null (vs undefined)? Could this be done on initialization and not on login?

u9g:
> Why null (vs undefined)? Could this be done on initialization and not on login?

nulls can be checked against and it's the standard in mineflayer, so I adopted it here

u9g:
this change will hopefully only be merged with the /clear one, because that one fixes the tests it breaks

extremeheat:
Okay, then I think this should be done in the createWindow call. Otherwise this should be done after all the other createWindow calls if not already. Else you get a mix of null and undefined.

Karang:
This would be best to to this in the constructor https://github.com/PrismarineJS/prismarine-windows/blob/master/lib/Window.js#L14
```js
this.slots = new Array(slotCount).fill(null)
```

u9g:
remade here: https://github.com/PrismarineJS/prismarine-windows/pull/22

PR #483, opened 2021-02-19
Add brigadier command dispatcher

u9g:


u9g:
Things to add:

- [ ] More targets (ie: entities, selectors, etc)
- [ ] More commands
- [ ] Add tests

Specific goals

- [ ] Add selector string support

rom1504:
old

PR #488, opened 2021-03-08
Unload chunks

Saiv46:
Yes, `flying-squid` keeps all chunks loaded until got Out-Of-Memory error

### DIFF
--- a/src/lib/plugins/logout.js
+++ b/src/lib/plugins/logout.js
@@ -19,6 +19,7 @@ module.exports.player = function (player, serv, { worldFolder }) {
 
   player._client.on('end', async () => {
     if (player && player.username) {
+      player.unloadAllChunks()
       serv.broadcast(serv.color.yellow + player.username + ' quit the game.')
       player._writeOthers('player_info', {
         action: 4,
--- a/src/lib/plugins/players.js
+++ b/src/lib/plugins/players.js
@@ -103,7 +103,7 @@ module.exports.server = function (serv, { version }) {
       const newItem = new Item(item, count)
 
       player.inventory.slots.forEach((e, i) => {
-        if (e === undefined) return
+        if (!e) return
         if (e.type === parseInt(newItem.type)) {
           e.count += parseInt(count)
           player.inventory.updateSlot(e.slot, e)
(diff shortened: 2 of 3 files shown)


rom1504:
what is the idea of the change ?

if you want to decrease world memory, one solution could be a LRU, but this should be happening in pworld, not in flying-squid

Saiv46:
> what is the idea of the change ?
The idea is to actually unload chunks when player is too far from it (not just remove a key from `player.loadedChunks
`), same when using `/changeworld`. Now we have a `server.chunksUsed` with chunk keys and counter of it - unloading chunks that is not used by calling corresponding method of pworld.

In pworld a LRU can be implemented (and I'll implement it someday), but that's optional.
This goal of PR is to bring some stability to the running instance of `flying-squid`.

rom1504:
You can't actually remove the chunks if one player get too far as other players may still need them.

So you need a strategy to decide when to remove chunks.
The more memory efficient solution would be to keep only the union of what players need now but it would increase compute (would need to regenerate or reread chunks more often)
A less memory efficient but nicer on compute solution is LRU, to keep the chunks a bit longer in case players move back.

rom1504:
It seems you implemented the first solution I'm mentioning.
Maybe you can put it under an option so people can choose ?

Saiv46:
> You can't actually remove the chunks if one player get too far as other players may still need them.
I implemented a counter for that. The only problem I see is when players in two dimensions are loading chunks at same position - server will keep both and unload only one.

> So you need a strategy to decide when to remove chunks.

> The more memory efficient solution would be to keep only the union of what players need now but it would increase compute (would need to regenerate or reread chunks more often)
Unlike java server, `flying-squid` is very fast at regenerating/rereading chunks (even at max (64 chunks) render distance).
> A less memory efficient but nicer on compute solution is LRU, to keep the chunks a bit longer in case players move back.
That should be implemented in pworld, also see: https://github.com/PrismarineJS/prismarine-design/issues/1

rom1504:
do you intend to finish this? a minor change is still missing

rom1504:
Thanks, merged in @KaffinPX PR

PR #492, opened 2021-03-27
Make @Saiv46 chunks with some improvements

KaffinPX:
It unloads chunks when necessary, it decreases ram and cpu usage.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -168,7 +168,7 @@
     - [Low level properties](#low-level-properties)
       - [player._client](#player_client)
     - [Low level methods](#low-level-methods-1)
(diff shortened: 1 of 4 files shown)


extremeheat (approved):
You could make the pr more descriptive, it's not clear what this does without looking at the code. Looks like you fixed unloading chunks in use by other players by adding a ref count.
Did you mean to remove `player.unloadChunk` ? Doesn't seem to have been documented. Other than that, LGTM.

KaffinPX:
wait ill look to docs.

rom1504 (approved):

PR #496, opened 2021-05-13
Removes needle

KaffinPX:
Removes needle from packages, replaces with good built-in alternative


rom1504:
this is worse afterwards, let's not do it
feel free to replace needle by node-fetch though, it a bit better

KaffinPX:
lets merge it to netheritejs/flying-squid :P

rom1504:
feel free

PR #511, opened 2021-06-29
Moved Dockerfiles and updated readme

TheBrenny:
Moved the Dockerfile (and co) from the deploy folder to the root. I don't know if this affects anything in terms of gitpod or publishing to docker, but I hope not.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -8,3 +8,4 @@
 **/.editorconfig
 **/.aws
 **/dist
+**/node_modules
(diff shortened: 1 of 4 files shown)


rom1504:
what is the purpose of the change? does it work?

TheBrenny:
It's unclear as to how to build the docker image while the Dockerfile isn't in the root folder, and from my trial and many errors, it wouldn't work as expected. Making this changed fixed my issues, and I assumed that with no documentation to state why these files weren't in the root directory, it'd be a better alternative.

PR #512, opened 2021-06-29
Added jshintrc to suppress semicolons

TheBrenny:
Exactly that. I added a `.jshintrc` initially to suppress missing semicolons, but also added statements for esversion and bigint.

Maybe the codebase should be refactored to include semicolons anyway? 🤔

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -1,0 +1,7 @@
+{
+    "asi": true,
+    "esversion": 8,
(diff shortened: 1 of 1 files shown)


u9g:
Thanks for contributing, however this project uses [`standard`](https://github.com/standard/standard) for all linting, and it's been decided against using semicolons.

PR #518, opened 2021-07-23
1.16.5 support

### LINKED ISSUE #447: 1.16.2 and 1.16.3 support
we have nmp support now
https://github.com/PrismarineJS/node-minecraft-protocol/pull/757#issuecomment-678520061 indications for mineflayer support
we need adaptations for use of these 3 packets
(cc @GroobleDierne )

GroobleDierne:
Close #447 
**Still work in progress.**

I have some questions: here https://github.com/PrismarineJS/flying-squid/blob/master/src/lib/version.js#L2
the supported versions are specified with their minor version but here https://github.com/PrismarineJS/flying-squid/blob/master/src/index.js#L46 the minor version is ignored. So I was wondering if I should change it to check also the minor version, so only the latest minor version will be supported for each Minecraft major version. That would simplify development (especially for 1.16 where there are a lot of differences between 1.16.1 and 1.16.2 and superior) but it will reduce the number of available versions.

Also, I didn't found a way to trigger a multi_block_change packet (here https://github.com/PrismarineJS/flying-squid/blob/master/src/lib/plugins/blockUpdates.js#L153)


u9g:
> Also, I didn't found a way to trigger a multi_block_change packet (here https://github.com/PrismarineJS/flying-squid/blob/master/src/lib/plugins/blockUpdates.js#L153)

What do you mean by you didn’t?

GroobleDierne:
> 
> 
> > Also, I didn't found a way to trigger a multi_block_change packet (here https://github.com/PrismarineJS/flying-squid/blob/master/src/lib/plugins/blockUpdates.js#L153)
> 
> What do you mean by you didn’t?

Flying-squid seems to send this packet if there is more than one block update at a tick (for example, an explosion), I didn't found a way to satisfies this condition just by breaking blocks in-game and, I was wondering if there is a /explosion command or a /fill command or any other way to make more than one update per tick in-game.

u9g:
How far along is this?

Saiv46:
@GroobleDierne Hi, can I remake this PR if you don't want to finish this?

GroobleDierne:
Feel free to do so, I won't have the time to finish it before long.

rom1504:
anyone feel like finishing this?
@darksunlight maybe ?

BasToTheMax:
- How far are you?
- What needs to be done to support 1.16.5?

GroobleDierne:
I don't have time to finish it but if someone does I can pull their changes into this branch

BasToTheMax:
> I don't have time to finish it but if someone does I can pull their changes into this branch

I understand that, but I need to know what changes need to be made until 1.16.5 works

GroobleDierne:
> > I don't have time to finish it but if someone does I can pull their changes into this branch
> 
> I understand that, but I need to know what changes need to be made until 1.16.5 works

The multi_blocks_change packet still needs to be tested, in theory I updated it but at the time that packet was never used by flying-squid so I wasn't able to test it.
Also there is still a pending question that I asked in the PR description, it's that I took the initiative of only supporting the latest minor release of each major minecraft release (so not supporting 1.16.0-1.16.4 but only 1.16.5) because I saw that albeit the list of supported versions contained every minor release for every major release, when checking if the server supports a specific version, the minor number is stripped and only the major is looked for (1.16 instead of 16.5), so we need to know if my initiative is good or not and revert it if so.

andriycraft:
#599

rom1504:
too old, feel free to redo

PR #519, opened 2021-08-29
Fix missing return for spawnObject

para-dise:
Fixes a missing return for spawnObject, spawnMob returns a mob and spawnObject should therefore return the object


u9g (approved):
LGTM, thanks

PR #549, opened 2022-01-10
Pin compromised color package

extremeheat:
Package is in bad hands and should be replaced, but for now the old version are locked by npm so should be a temporary fix


rom1504:
want to do a release pr ?

PR #551, opened 2022-01-18
Fix broken link in README.md

ghost:
I removed the link to autonomous-squid
Because the repository doesn't exists


andriycraft:
Moved to #554

PR #552, opened 2022-01-18
Create maxplayers.js

andriycraft:


andriycraft:
Done

rom1504:
please add doc for this new option

andriycraft:
@rom1504

src/lib/plugins/maxplayers.js
```diff
@@ -0,0 +1,16 @@
+let playersonline = 0
```
rom1504, line 1: Can you use a variable of the server instead of a global ?

andriycraft:
@rom1504

src/lib/plugins/maxplayers.js
```diff
@@ -0,0 +1,16 @@
+module.exports.player = async function (player, serv, settings) {
+  serv.playersonline = 0
```
rom1504, line 2: that will reset the number of players whenever a new player joins

rom1504:
you could use the existing players object to check the number of players

PR #555, opened 2022-01-21
Create uncaughtException.js

andriycraft:


rom1504:
what is this

andriycraft:
> what is this

process.on('uncaughtException'...) will trigger when code occurs error
process.on('uncaughtException'...) will trigger when code occurs unhandled rejection

[https://www.geeksforgeeks.org/node-js-process-uncaughtexception-event/](url)

For example:
if i do:
  player.on('connected', () => {
   some_function_that_doesnt_exists()
  }
Then the server will not crash, and will log the information of the error in console

rom1504:
this will keep the server running in case of issues, I'm not sure if that's good

PR #559, opened 2022-02-01
Fix a duplication problem that causes fake blocks

KaffinPX:


src/lib/plugins/blocks.js
```diff
@@ -2,6 +2,8 @@ const Vec3 = require('vec3').Vec3
 
 module.exports.player = function (player, serv) {
   player.changeBlock = async (position, blockType, blockData) => {
+    if (Math.round(player.gameMode) === 1 && player.heldItem?.name.endsWith('_sword')) return
```
rom1504, line 5: why

KaffinPX: You can't break blocks with sword on creative mode, idk why gameMode has decimals

rom1504:
This pull request makes 0 sense
changeBlock shouldn't be called in the first place when the user is calling a sword, so it's the wrong place to make this change

gameMode should not be rounded

Please fix or close

PR #560, opened 2022-02-06
Fixed opening of chest GUI window and chest size

AlexNemo:


config/default-settings.json
```diff
@@ -2,7 +2,7 @@
   "motd": "A Minecraft Server \nRunning flying-squid",
   "port": 25565,
   "max-players": 10,
-  "online-mode": true,
+  "online-mode": false,
```
rom1504, line 5: Why this change?

AlexNemo: Oops... Sorry about it... I forgot that I've changed this file... It was done for the testing purpose and actually is not intended to go in the repository. If you wish it to be like it was in the repository before you can either change it yourself or I will fix it in the next commit

src/lib/plugins/chest.js
```diff
@@ -1,26 +1,103 @@
...
+        player.windowType = 'shulker_box'
+      }
+      player.windowPos = block.position
+      // Opening chest GUI window
+      player._client.write('open_window', {
+        windowId: player.windowId,
+        inventoryType: 2,
+        windowTitle: JSON.stringify(chestWindowTitle)
+      })
+      // Sending chest content (NOT IMPLEMENTED)
```
rom1504, line 66: Looks like sending the chunk content is not implemented

src/lib/plugins/chest.js
```diff
@@ -1,26 +1,103 @@
...
+      // Getting current block and block above it
+      const id = await player.world.getBlockType(block.position)
+      const blockAbove = await player.world.getBlockType(block.position.plus(new Vec3(0, 1, 0)))
+      // Getting ALL supported blocks
+      // Chests
+      const blockChest = mcData.blocksByName.chest
+      const blockEnderChest = mcData.blocksByName.ender_chest
+      // TODO: Large chest (NOT IMPLEMENTED)
+      // Shulker boxes
+      const blockShulkerBox = mcData.blocksByName.shulker_box
```
rom1504, line 18: Let's use a list for this instead

src/lib/plugins/chest.js
AlexNemo: Yes, it isn't. I can't implement it now because if the content isn't actually stored then I have nothing to send

src/lib/plugins/chest.js
```diff
@@ -1,26 +1,89 @@
...
     }
-  })
+  }
+  // Registering block interaction handlers
+  // Chests
+  serv.onBlockInteraction(mcData.blocksByName.chest.name, chestsBlockInteractionHandler)
+  serv.onBlockInteraction(mcData.blocksByName.ender_chest.name, chestsBlockInteractionHandler)
+  // Shulker boxes
+  if (serv.supportFeature('theShulkerBoxes')) {
+    serv.onBlockInteraction(mcData.blocksByName.shulker_box.name, chestsBlockInteractionHandler)
```
rom1504, line 71: can you use a loop and the list there ?

src/lib/plugins/chest.js
rom1504: it can be stored in the world in memory, it doesn't have to be stored on the disk

src/lib/plugins/chest.js
AlexNemo: There too? Ok. Will do it

rom1504:
Code looks good now.

For the chest feature, missing:
* [ ] Storing the chest state in memory
* [ ] Storing the chest state on disk
* [ ] Expanding the tests to test the feature works


In this PR, can you do the testing part (there https://github.com/PrismarineJS/flying-squid/blob/master/test/mineflayer.test.js#L150 ? ) ?

AlexNemo:
What exactly do I need to test?

rom1504:
All the features that this PR is adding
From what I see there is :
* Opening chest
* Opening shulker box
Is there anything else ?

The objective of the tests is that if someone changes the code and break the features you are adding them the test will fail and we won't merge the changes

AlexNemo:
Okay. How can I test it? As far as I understand Mineflyer can't actually open GUIs, change slots of said guis like player does and then after cosing GUI open it again and check if the chest/shulker box slots are saved

rom1504:
mineflayer can do all that

rom1504:
looks correct so merging
my point above regarding finishing the chest feature remain

PR #563, opened 2022-02-13
Fix drops

darksunlight:
- adds basic support for enchantments
- fixes item drops (visually) (1.15+)
- implements more accurate drops based on tools (types, enchantments) used (1.14+)

### DIFF
--- a/src/lib/playerDat.js
+++ b/src/lib/playerDat.js
@@ -34,7 +34,7 @@ async function read (uuid, spawnPoint, worldFolder) {
         pitch: playerData.Rotation.value.value[1],
         onGround: Boolean(playerData.OnGround.value)
       },
-      inventory: playerData.Inventory.value.value
+      inventory: playerData.Inventory.value.value.map(nbtItem => { nbtItem.tag ? nbtItem.tag.name = '' : null ; return nbtItem})
     }
   } catch (e) {
     return {
@@ -44,12 +44,12 @@ async function read (uuid, spawnPoint, worldFolder) {
   }
 }
 
-async function save (player, worldFolder, snakeCase) {
+async function save (player, worldFolder, snakeCase, theFlattening) {
   function playerInventoryToNBT (playerInventory) {
     const nbtInventory = []
     playerInventory.slots.forEach(item => {
       if (item) {
-        nbtInventory.push({
+        const nbtItem = {
           Slot: {
             type: 'byte',
             value: convertInventorySlotId.toNBT(item.slot)
@@ -61,12 +61,21 @@ async function save (player, worldFolder, snakeCase) {
           Count: {
             type: 'byte',
             value: item.count
-          },
-          Damage: {
-            type: 'short',
-            value: item.metadata
           }
-        })
+        }
+        if (!theFlattening) {
+          Object.assign(nbtItem, {
+            Damage: {
+              type: 'short',
+              value: item.metadata
+            }
+          })
+        } else if (item.nbt) {
+          Object.assign(nbtItem, {
+            tag: item.nbt
+          })
+        }
+        nbtInventory.push(nbtItem)
       }
     })
     return nbtInventory
--- a/src/lib/playerDat.js
+++ b/src/lib/playerDat.js
@@ -34,7 +34,10 @@ async function read (uuid, spawnPoint, worldFolder) {
         pitch: playerData.Rotation.value.value[1],
         onGround: Boolean(playerData.OnGround.value)
       },
-      inventory: playerData.Inventory.value.value.map(nbtItem => { nbtItem.tag ? nbtItem.tag.name = '' : null ; return nbtItem})
+      inventory: playerData.Inventory.value.value.map(nbtItem => {
+        if (nbtItem.tag) nbtItem.tag.name = ''
+        return nbtItem
+      })
     }
   } catch (e) {
     return {
--- a/src/lib/playerDat.js
+++ b/src/lib/playerDat.js
@@ -35,7 +35,7 @@ async function read (uuid, spawnPoint, worldFolder) {
         onGround: Boolean(playerData.OnGround.value)
       },
       inventory: playerData.Inventory.value.value.map(nbtItem => {
-        if (nbtItem.tag) nbtItem.tag.name = ''
+        if (nbtItem.tag && nbtItem.tag === undefined) nbtItem.tag.name = ''
         return nbtItem
       })
     }
--- a/src/lib/plugins/commands.js
+++ b/src/lib/plugins/commands.js
@@ -205,8 +205,8 @@ module.exports.server = function (serv, { version }) {
 
   const notudf = i => typeof i !== 'undefined'
 
-  serv.selector = (type, opt) => {
-    if (['all', 'random', 'near', 'entity'].indexOf(type) === -1) { throw new UserError('serv.selector(): type must be either [all, random, near, or entity]') }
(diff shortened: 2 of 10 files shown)


rom1504:
what's the status here ?

darksunlight:
> what's the status here ?

I just have to
- replace my temporary selector parser
- implement check for version for entity metadata
- implement check for enchantment max level and compatibility

andriycraft (approved):

src/lib/plugins/players.js
```diff
@@ -88,36 +103,70 @@ module.exports.server = function (serv, { version }) {
...
+    info: 'Enchants items holded by targets with the specified enchantment and level',
+    usage: '/enchant <targets> <enchantment> [level]',
+    tab: ['selector', 'item_enchantment', 'number'],
+    op: true,
+    parse (args, ctx) {
+      args = args.split(' ')
+      if (args[0] === '') return false
+      const enchantment = mcData.enchantmentsByName[args[1]]
+      if (!enchantment) throw new UserError('No such enchantment')
+      if (args[2] && (parseInt(args[2]) > enchantment.maxLevel || parseInt(args[2]) < 0)) throw new UserError('Level invalid')
```
andriycraft, line 153: Level is invalid*

src/lib/plugins/players.js
```diff
@@ -32,7 +37,7 @@ module.exports.server = function (serv, { version }) {
         else throw new UserError('Console cannot set gamemode itself')
       }
 
-      return str.match(/^(survival|creative|adventure|spectator|[0-3]) (\w+)$/) || false
+      return str.match(/^(survival|creative|adventure|spectator|[0-3])( @[arspe](?:\[([^\]]+)\])?| \w+)?$/) || false
       // return params || false
```
andriycraft, line 41: Remove this comment

andriycraft (changes requested):

src/lib/plugins/players.js
darksunlight: Why? I think having a descriptive comment of the cryptic regex is useful.

andriycraft: Ok

andriycraft (approved):

src/lib/playerDat.js
```diff
@@ -34,7 +34,10 @@ async function read (uuid, spawnPoint, worldFolder) {
         pitch: playerData.Rotation.value.value[1],
         onGround: Boolean(playerData.OnGround.value)
       },
-      inventory: playerData.Inventory.value.value
+      inventory: playerData.Inventory.value.value.map(nbtItem => {
+        if (nbtItem.tag) nbtItem.tag.name = ''
```
rom1504, line 38: Why

src/lib/plugins/placeBlock.js
```diff
@@ -16,7 +16,7 @@ module.exports.server = (serv, { version }) => {
   const itemPlaceHandlers = new Map()
   serv.placeItem = (data) => {
     const handler = itemPlaceHandlers.get(data.item.type)
-    return handler ? handler(data) : { id: data.item.type, data: data.item.metadata }
+    return handler ? handler(data) : (serv.supportFeature('theFlattening') ? {} : { id: data.item.type, data: data.item.metadata })
```
rom1504, line 19: This is wrong

rom1504:
Sorry misclicked

rom1504:
This PR is a collection of random stuff, some seems clearly good and some not obvious
What do you think about opening some more specific PRs instead?

That or explaining more what is each of the 10 changes supposed to do ?

src/lib/plugins/placeBlock.js
darksunlight: I don't think I understand what you mean.

src/lib/playerDat.js
darksunlight: nbtItem.tag.name would be undefined otherwise and would cause an error on join.

darksunlight:
@rom1504 
Here's a breakdown of the changes:

| File name | Description |
| ---------- | ----------- |
| package.json | increase mocha timeout to resolve randomly failing CI tests and bump minecraft-data for up-to-date data |
| playerDat.js | changes to player.dat saving and retrieving to support enchantments |
| commands.js | addition of @s selector (for manual testing with /enchant command) |
| digging.js | fix item drops |
| login.js | read item data properly for 1.13+ to support NBT (including enchantments) |
| logout.js | for changes in playerDat.js |
| placeBlock.js | prevent items with no corresponding blocks to be placed (this bug annoyed me a lot when testing item drops) |
| players.js | 1. added convenient function to get players given a selector string for use in /enchant<br>2. Fix /gamemode command (annoying bug that hindered testing)<br>3. Change /give to use the new convenient function<br>4. Add the /enchant command |
| spawn.js | spawn items with correct metadata thus fixing their visual appearance |
| tabComplete.js | misc changes to facilitate testing |

src/lib/plugins/placeBlock.js
rom1504: why would we return nothing after the flattening ?

src/lib/playerDat.js
rom1504: if it's defined and not empty you're losing the value
but it sounds like the problem should be fixed wherever the error was happening instead of here

src/lib/plugins/placeBlock.js
darksunlight: placeItem is currently only used in the same file, handling the block_place packet from clients. It's destructured into id and data after being called, then the id is used to find the corresponding blocks. Blocks with the same ID as an item is not the corresponding block for the item when the item does not have a corresponding block, so I return an empty object with ID undefined so the wrong block won't be placed.

darksunlight: In pre-flattening times this can be used as a fallback for items that somehow don't have an item place handler as the old assumption is indeed true.

src/lib/playerDat.js
darksunlight: nbtItem.tag.name is intended to be empty. It's just somehow lost during the saving process, and I don't know prismarine-nbt enough to figure out what caused it to omit this empty string value.

andriycraft (approved):

src/lib/playerDat.js
rom1504: can you do if (nbtItem.tag && nbtItem.tag == undefined)

rom1504: did it

src/lib/plugins/commands.js
```diff
@@ -215,6 +215,7 @@ module.exports.server = function (serv, { version }) {
     const pos = opt.pos
     let sample
     if (type === 'all') sample = serv.players
+    else if (type === 'self') sample = serv.players.filter(p => p.id === selfEntityId)
```
rom1504, line 218: this needs to be added in doc

src/lib/plugins/commands.js
```diff
@@ -287,17 +288,18 @@ module.exports.server = function (serv, { version }) {
     else return sample.slice(count) // Negative, returns from end
   }
 
-  serv.selectorString = (str, pos, world, allowUser = true) => {
+  serv.selectorString = (str, pos, world, allowUser = true, ctxEntityId) => {
```
rom1504, line 291: add in doc

rom1504:
I'm going to merge.

If we get reports of things being broken due to that, I will revert, and the next steps will be to create many small PRs doing one unitary changes for this instead of this massive PR.

PR #576, opened 2022-03-14
[/tp command] - Check if x/y/z is valid

andriycraft:


src/lib/plugins/tp.js
```diff
@@ -25,6 +25,11 @@ module.exports.server = (serv) => {
         let y = serv.posFromString(args[1], ctx.player.position.y)
         let z = serv.posFromString(args[2], ctx.player.position.z)
 
+        if (x > 29999999 || y > 4096 || y < -4096 || z > 29999999) {
```
darksunlight, line 28: What about negative x and z coordinates? You can check for the absolute values of x y and z instead.

darksunlight (changes requested):

src/lib/plugins/tp.js
```diff
@@ -25,6 +25,11 @@ module.exports.server = (serv) => {
         let y = serv.posFromString(args[1], ctx.player.position.y)
         let z = serv.posFromString(args[2], ctx.player.position.z)
 
+        if (x > 29999999 || y > 4096 || y < -4096 || z > 29999999 || z < -29999999 || x < -29999999) {
```
MokiyCodes, line 28: using `Math.abs(x) > 29999999 || Math.abs(y) > 4096 || Math.abs(z) > 29999999` would likely look nicer

MokiyCodes (changes requested):
using `Math.abs(x) > 29999999 || Math.abs(y) > 4096 || Math.abs(z) > 29999999` would likely look nicer

andriycraft:
> using `Math.abs(x) > 29999999 || Math.abs(y) > 4096 || Math.abs(z) > 29999999` would likely look nicer

Remember that we also have for negative x/y/z
so your code will look like this
`Math.abs(x) > 29999999 || Math.abs(y) > 4096 || Math.abs(z) > 29999999 || Math.abs(x) < -29999999 || Math.abs(y) < -4096 || Math.abs(z) < -29999999`

so I think my code is nicer

andriy332 (approved):

darksunlight:
> > using `Math.abs(x) > 29999999 || Math.abs(y) > 4096 || Math.abs(z) > 29999999` would likely look nicer
> 
> Remember that we also must have check for negative x/y/z so your code will look like this `Math.abs(x) > 29999999 || Math.abs(y) > 4096 || Math.abs(z) > 29999999 || Math.abs(x) < -29999999 || Math.abs(y) < -4096 || Math.abs(z) < -29999999`
> 
> so I think my code is nicer

No. Not really. Math.abs returns the absolute value of its first parameter. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/abs

MokiyCodes (approved):
lgtm

Issue #323, opened 2018-02-16
Flow/TypeScript


mhsjlw:
Explore these tools

demipixel:
Also, we need to decide what bundler to use. Webpack is popular but I'm not sure if it's the best for the job.

rom1504:
Why do we need a bundler ? I thought the types were just for the ide to autocomplete stuff and for tools to check correct typing.a

demipixel:
Well it's not regular JS, so at the very least you need to compile the ts files into js ones (or a single js one).

rom1504:
Why ? I don't understand.
What are we going to do with a resulting js ?

demipixel:
You can't run TS files, you have to convert them to JS files and run those (although using stuff like watch you can automate that).

rom1504:
But we don't want to run any ts files ?

We just want to have some autocompletion right ?
I still don't understand.

demipixel:
Well we were talking about converting the project to TypeScript. Manually handling the typings means any time you change something in the API, you have to remember to change it in the typings.

rom1504:
No I don't think we should convert the project to typescript.
I thought typescript types could be used along with JavaScript code.

demipixel:
Well, you can write a typings file and manually handle it from that. In addition, flow provides the option to add commented typings.

mhsjlw:
TypeScript just has it's own `tsc` command that (might?) work fine. Although we may want to use something like webpack so we can set up a development and production profile.

rom1504:
I don't think we should convert PrismarineJS stuff to typescript.
Adding types in separate files, ok, but if we need to convert the code files to ts I don't think it's a good idea.

mhsjlw:
No, the prismarine modules do not need to be converted.

mhsjlw:
Remember, though, that we don't _need_ to use TypeScript, we could always use Flow and flow-remove-types which just simply removes the typings from the file producing the same JavaScript. As well, we can convert TypeScript typings to flow definitions https://github.com/bcherny/flow-to-typescript and by nature, Flow is less intrusive than TypeScript as everything doesn't need to be converted and you can just omit types if you wish.

mhsjlw:
Also, we can just include the typings for prismarine inside of the flying-squid repository as that is normal for most flow projects to do.

rom1504:
so even with flow we would need to compile files ?

mhsjlw:
[...]

PR #579, opened 2022-03-20
Move to Typescript

### LINKED ISSUE #533: TypeScript Support
What about TypeScript support? Could not find anything about this and could not install any Types. It is a great support in the development and helps to prevent mistakes and better understand what options you have.

MokiyCodes:
## Changes
This PR introduces a few changes:
1. Move to TypeScript (duh) - This also means there are now auto-generated typescript definitions
2. Make test cases use node 14.x and 17.x, rather than just 14.x
3. Use eslint directly; standard complains about typescript's syntax
4. Update Dependencies
5. Add names to each CI/CD step
6. Move linting to a seperate workflow, to avoid unnecesary linting for each node version
7. Add TypeScript `examples/basic.ts`
8. Make `package.json`:`engines.node` be `>=14` instead of the previous `>=8`

In terms of TypeScript types, there's a lot more to be desired, however this is definitely an improvement over vanilla JS.

## Compatability
This, in theory, shouldn't break any dependents, unless they directly import files in the `src` directory.

## Other things of note
[src/lib/plugins/log.ts:49](https://github.com/MokiyCodes/flying-squid/blob/47e9ccde9c9292830518a484ce59433c7ce122cd/src/lib/plugins/log.ts#L49) Removed this weird code that my tsc was complaining about - don't really know why it's there, maybe someone here can re-implement this in a non-broken way.

---

[Conventional Commits](https://conventionalcommits.org) [Fixes #533](https://github.com/PrismarineJS/flying-squid/issues/533)


MokiyCodes:
The CI runs here won't succeed, considering these don't run `npm run build`

MokiyCodes:
This also fixes #323

rom1504:
can you instead follow the same technical choice we made for the rest of PrismarineJS: adding a .d.ts file and keeping the code in javascript?

MokiyCodes:
> can you instead follow the same technical choice we made for the rest of PrismarineJS: adding a .d.ts file and keeping the code in javascript?

Sure - I'll do that when I have time for that.

IceTank:
> [src/lib/plugins/log.ts:49](https://github.com/MokiyCodes/flying-squid/blob/47e9ccde9c9292830518a484ce59433c7ce122cd/src/lib/plugins/log.ts#L49) Removed this weird code that my tsc was complaining about - don't really know why it's there, maybe someone here can re-implement this in a non-broken way.
This looks like it is patching console.log in a way where it will always put the cursor at a certain position once called. It is probably to make the command line cleaner once stuff has been logged to it. Just like the vanilla minecraft server dose it. No wonder it is throwing linting errors because it overwrites the default console.log. The 'real' way to patch this is to make a new custom logging function and change all console.log to that.

Bezouro:
> can you instead follow the same technical choice we made for the rest of PrismarineJS: adding a .d.ts file and keeping the code in javascript?

@rom1504 
Doesn't this mean more work to maintain the project?
Why go this way when we can have the code already in typescript?

PR #595, opened 2022-05-20
Fix breaking/placing blocks in adventure mode

Saiv46:


kf106:
I've tried this, but I'm finding that I can still destroy blocks, albeit rather slowly.

Placing block appears to be prevented.

kf106:
Looks like L63-L64 are calling completeDigging which allows a block to be destroyed:

I'll try with Badlion to see if somehow prismarine-web-client is sneaking past the code.

Saiv46:
> I've tried this, but I'm finding that I can still destroy blocks, albeit rather slowly. I'm using flying squid in 1.15.2 with prismarine web client in 1.15.2.

Prismarine Web Client seems to ignore absence of `acknowledge_player_digging` packet (see PrismarineJS/mineflayer#1555) and continues to show breaking animation.

Now server will cancel digging unconditionally.

kf106:
Hmm, I don't know what's going on here. I can still break blocks using Prismarine, and then walk through the gap that is made.

Saiv46:
Web client is based on Mineflayer, which seems to ignore such things as server-side dig cancelling.

Can you please test it via vanilla client?

kf106:
@Saiv46 - This works in Badlion, but not in Prismarine Viewer.

I suddenly had a thought, and broke blocks, moved away to trigger a chunk unload, and moved back. The blocks had magically reappeared. So the digging is not saved or taking effect in Flying Squid, but it is in PV.

I then ran two PV web pages next to each other. The second player does not see the broken blocks. But ... the first player can break blocks and walk into the gap that is made. The second player then sees this:
embedded

Saiv46:
@rom1504 Some PRs (except 1.16.5 support) are good to go

PR #596, opened 2022-05-21
Commands(OP): Make player non-case-sensitive

u9g:
/op u9g should still find U9G, now it does :)

### DIFF
--- a/src/lib/plugins/moderation.js
+++ b/src/lib/plugins/moderation.js
@@ -84,8 +84,9 @@ module.exports.server = function (serv, settings) {
       return params
     },
     action (params) {
-      params = params.split(' ')
-      const player = serv.getPlayer(params[0])
+      const [inputUsername] = params.split(' ')
+      // get player, by non-case-sensitive username
+      const player = serv.players.find(player => player.username.toLowerCase() === inputUsername.toLowerCase())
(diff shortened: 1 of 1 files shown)


src/lib/plugins/moderation.js
```diff
@@ -98,10 +99,10 @@ module.exports.server = function (serv, settings) {
         if (!player.op) {
           player.op = true
 
-          player.chat(`§7§o[Server: Opped ${params[0]}]`)
-          return `Opped ${params[0]}`
+          player.chat(`§7§o[Server: Opped ${player.username}]`)
```
andriycraft, line 102: Why [Server:.. ?
why not ${player.username || 'Server'

andriycraft (changes requested):

src/lib/plugins/moderation.js
u9g: done in latest commit

andriycraft (approved):

andriycraft (changes requested):
You forgot to make /deop command non case sensitive

andriycraft:
@rom1504 what will happen if there are 2 players on the server: 'Kinash___' and 'KINASH___' then what player will be opped if i use /op Kotinash___

PR #597, opened 2022-05-25
Improve error message

andriycraft:


BasToTheMax:
You currently have `You can download new version from...`, but you should make it:
`You can download the new version from...`

PR #603, opened 2022-08-11
Fixes to moderation.js + stuff

andriycraft:


andriycraft:
@rom1504

config/default-settings.json
```diff
@@ -1,12 +1,12 @@
...
+  "motd": "A Minecraft Server \nRunning Flying Squid",
   "port": 25565,
-  "max-players": 10,
+  "max-players": 25,
   "online-mode": true,
   "logging": true,
-  "gameMode": 1,
-  "difficulty": 1,
-  "worldFolder":"world",
+  "gameMode": 0,
```
darksunlight, line 7: Survival mode is practically unplayable with flying-squid currently.

src/lib/plugins/moderation.js
```diff
@@ -16,7 +16,7 @@ module.exports.server = function (serv, settings) {
     if (!serv.bannedIPs[IP]) {
       serv.bannedIPs[IP] = {
         time: +moment(),
-        reason: reason || 'Your IP is banned!'
+        reason: reason || 'You has been IP banned from this server. Bye!'
```
darksunlight, line 19: A user seeing this message might not necessarily be the intended target of the IP ban, but is framed as such. The original message is more factually and grammatically correct.

src/lib/plugins/moderation.js
```diff
@@ -119,7 +119,9 @@ module.exports.server = function (serv, settings) {
     },
     action (params) {
       params = params.split(' ')
-      const player = serv.getPlayer(params[0])
+      const [inputUsername] = params.split(' ')
+      // get player, by non-case-sensitive username
+      const player = serv.players.find(player => player.username.toLowerCase() === inputUsername.toLowerCase())
```
darksunlight, line 124: It's probably better to modify serv.getPlayer instead since this check would be useful for other commands as well.

src/lib/plugins/moderation.js
```diff
@@ -132,7 +134,7 @@ module.exports.server = function (serv, settings) {
         if (player.op) {
           player.op = false
 
-          player.chat(`§7§o[Server: Deopped ${params[0]}]`)
+          player.chat(`§7§o[${player.username ?? 'Server'}: Deopped ${params[0]}]`)
```
darksunlight, line 137: It might be unfavourable for the deopped target to know who the deopper is.

src/lib/plugins/moderation.js
andriycraft: I changed it to the vanilla minecraft kick message

config/default-settings.json
andriycraft: Fixed

src/lib/plugins/moderation.js
andriycraft: I copied the code from /op command 
So i guess its ok

src/lib/plugins/moderation.js
andriycraft: Ok

andriycraft:
@watchers

andriycraft:
@rom1504

src/lib/plugins/moderation.js
```diff
@@ -16,7 +16,7 @@ module.exports.server = function (serv, settings) {
     if (!serv.bannedIPs[IP]) {
       serv.bannedIPs[IP] = {
         time: +moment(),
-        reason: reason || 'Your IP is banned!'
+        reason: reason || 'You has been IP banned from this server'
```
u9g, line 19: This isn’t correct grammer. The correct way to say this is: “You have been…”

src/lib/plugins/moderation.js
```diff
@@ -33,9 +33,9 @@ module.exports.server = function (serv, settings) {
     return await new Promise((resolve, reject) => {
       needle('get', 'https://api.mojang.com/users/profiles/minecraft/' + username, { json: true })
         .then((response) => {
-          if (!response.body) throw new Error('username not found')
+          if (!response.body) throw new Error('Invalid player!')
```
u9g, line 36: Why is it better to make the error more vague?

src/lib/plugins/moderation.js
```diff
@@ -132,7 +134,7 @@ module.exports.server = function (serv, settings) {
         if (player.op) {
           player.op = false
 
-          player.chat(`§7§o[Server: Deopped ${params[0]}]`)
+          player.chat('§cYou has been deopped')
```
u9g, line 137: What was wrong with it before?

rom1504:
will you finish this ?

andriycraft:
> will you finish this ?

Done

src/lib/plugins/moderation.js
```diff
@@ -119,7 +119,9 @@ module.exports.server = function (serv, settings) {
     },
     action (params) {
       params = params.split(' ')
-      const player = serv.getPlayer(params[0])
+      const [inputUsername] = params.split(' ')
```
u9g, line 122: this assumes inputUsername was found and is a string, maybe you can check before so it won't throw if the result of .split is null?

somucheffort: params is already split, why did you write another split?

PR #604, opened 2022-08-13
Fixed two block placement bugs

TheoPierne:
Fixed two bugs:

- A bug which allowed to place a block on the same position as the player 
- A bug which duplicated the block which was placed on a grass block

And add the direction on the block object to perform some actions with `serv.interactWithBlock`

### DIFF
--- a/src/lib/plugins/placeBlock.js
+++ b/src/lib/plugins/placeBlock.js
@@ -108,13 +108,15 @@ module.exports.player = function (player, serv, { version }) {
     const referencePosition = new Vec3(location.x, location.y, location.z)
     const block = await player.world.getBlock(referencePosition)
     block.position = referencePosition
+    block.direction = direction
     if (await serv.interactWithBlock({ block, player })) return
 
     const heldItem = player.inventory.slots[36 + player.heldItemSlot]
     if (!heldItem || direction === -1 || heldItem.type === -1) return
 
-    const directionVector = directionToVector[direction]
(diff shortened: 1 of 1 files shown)


andriycraft (approved):
approved

src/lib/plugins/placeBlock.js
```diff
@@ -108,14 +108,16 @@ module.exports.player = function (player, serv, { version }) {
...
     block.position = referencePosition
+    block.direction = direction
     if (await serv.interactWithBlock({ block, player })) return
     if (player.gameMode >= 2) return
 
     const heldItem = player.inventory.slots[36 + player.heldItemSlot]
     if (!heldItem || direction === -1 || heldItem.type === -1) return
 
-    const directionVector = directionToVector[direction]
+    const directionVector = block.name === 'grass' ? new Vec3(0, 0, 0) : directionToVector[direction]
```
darksunlight, line 118: There are only six directions. Why are you hardcoding this inexistent direction for grass?

TheoPierne: Currently if you place a block on a block of grass, the block you have placed will position itself relative to the face of the block of grass that you are aiming for with your crosshair. But in the Vanilla version of Minecraft, if you place a block on a grass block, the grass block is removed to be replaced by the block you want to place.

(The variable "block", is not the block you actually try to place by the way, is the block that you are aiming to place the block that you have in your hand)

And with the current version of the code (master branch), it also creates a bug where the grass block will disappear and be replaced by a ghost block of the same type as the block you wanted to put down.

PR #633, opened 2023-09-03
Refactor: allow usage in browser

### LINKED ISSUE #491: [feat req] Use in browser via webpack, browserify
Hello,
It would be nice if flying-squid would be webpackable to use completely client-side.
What currently limits flying-squid from being webpacked, and can these limitations be ignored to create a simple server?

zardoy:
Depends on https://github.com/PrismarineJS/node-minecraft-protocol/pull/1254

fixes #491

Usage:

Both esbuild and webpack needs Node.js modules polyfilling.
- esbuild: needs custom plugin like [this one](https://github.com/zardoy/prismarine-web-client/blob/77de78a92b4d52b1b039fb2ee9fb47a761da1865/scripts/esbuildPlugins.mjs#L157-L184)
- webpack: needs `module: { parser: { javascript: { commonjsMagicComments: true } } }`

First of all, I'm really sorry for the delay, I wish I could do it early.

Anyway, right now I decided to squash all changes [from here](https://github.com/PrismarineJS/flying-squid/compare/master...zardoy:space-squid:master) into one commit.

Will try to provide a brief overview of the changes I made here in comments.

- Refactored plugin loading so its compatible with webpack (at least I tested the first version of it) and esbuild that I use currently.

- I did initially a huge refactor by adding types, but I reverted them so its easier to review. Anyway, feel free to ask any questions


package.json
```diff
@@ -21,7 +19,11 @@
     "test": "npm run mocha_test",
     "pretest": "npm run lint"
   },
-  "keywords": [],
+  "keywords": [
```
zardoy, line 22: FYI just added some keywords (why not)

src/index.js
```diff
@@ -38,23 +38,23 @@ function createMCServer (options) {
 
 class MCServer extends EventEmitter {
   constructor () {
+    plugins.initPlugins()
     super()
     this._server = null
   }
 
   connect (options) {
+    this.options = options
```
zardoy, line 47: Not sure why it wasn't exposed before, but I need to change options later (that are used in plugins).

src/index.js
```diff
@@ -38,23 +38,23 @@ function createMCServer (options) {
 
 class MCServer extends EventEmitter {
   constructor () {
+    plugins.initPlugins()
```
zardoy, line 41: we now load plugins lazily

src/lib/plugins/log.js
```diff
@@ -1,20 +1,39 @@
...
+const _servers = []
+
+/** @type {typeof import("readline") | undefined} */
+let readline
+/** @type {import("readline").Interface | undefined} */
+let rl
+if (isInNode) {
+  import(/* webpackIgnore: true */ 'exit-hook').then((hook) => {
+    hook.default(() => {
+      for (const serv of _servers) {
```
zardoy, line 18: In the context of this refactor also fixed an important bug: worlds are not saved when server crashed or stopped by ctrl+c

src/lib/plugins/log.js
```diff
@@ -25,8 +44,9 @@ module.exports.server = function (serv, settings) {
   const logFile = path.join('logs', timeStarted + '.log')
 
   serv.log = message => {
-    readline.cursorTo(process.stdout, 0)
+    readline?.cursorTo(process.stdout, 0)
     message = moment().format('MMMM Do YYYY, HH:mm:ss') + ' ' + message
+    if (!isInNode) message = '[server] ' + message
```
zardoy, line 49: I think it always makes sense to do this when flying-squid is used as a dependency since there are other parts of an application that can do logging.

src/lib/plugins/log.js
```diff
@@ -46,25 +66,30 @@ module.exports.server = function (serv, settings) {
...
       }
-      rl.prompt(true)
-    }
-  })()
+    })()
+  }
 
   serv.createLog = () => {
     if (!settings.logging) return
-    mkdirp('logs', (err) => {
```
zardoy, line 67: remove mkdirp dep

src/lib/plugins/login.js
```diff
@@ -1,28 +1,26 @@
...
 const convertInventorySlotId = require('../convertInventorySlotId')
+const plugins = require('./index')
 
 module.exports.server = function (serv, options) {
   serv._server.on('connection', client =>
     client.on('error', error => serv.emit('clientError', client, error)))
 
   serv._server.on('login', async (client) => {
-    if (client.socket.listeners('end').length === 0) return // TODO: should be fixed properly in nmp instead
+    if (client.socket?.listeners('end').length === 0) return // TODO: should be fixed properly in nmp instead
```
zardoy, line 14: socket is not defined when using a custom server: like in https://github.com/zardoy/prismarine-web-client/blob/318847186b76882989abc996527804d15785635d/src/customServer.js (don't see a reason to mock it instead)

src/lib/plugins/login.js
```diff
@@ -1,28 +1,26 @@
...
+    if (!serv.worldsReady) throw new Error('World is still preparing.')
     try {
       const player = serv.initEntity('player', null, serv.overworld, new Vec3(0, 0, 0))
       player._client = client
+      player._client.socket ??= { remoteAddress: '' }
```
zardoy, line 19: same here

src/lib/plugins/login.js
```diff
@@ -1,28 +1,26 @@
...
+const plugins = require('./index')
 
 module.exports.server = function (serv, options) {
   serv._server.on('connection', client =>
     client.on('error', error => serv.emit('clientError', client, error)))
 
   serv._server.on('login', async (client) => {
-    if (client.socket.listeners('end').length === 0) return // TODO: should be fixed properly in nmp instead
+    if (client.socket?.listeners('end').length === 0) return // TODO: should be fixed properly in nmp instead
+    if (!serv.worldsReady) throw new Error('World is still preparing.')
```
zardoy, line 15: As I remember this is another bugfix: when a player connects when world is not ready yet it crashes

src/lib/plugins/login.js
```diff
@@ -66,13 +64,8 @@ module.exports.player = async function (player, serv, settings) {
...
     playerData.inventory.forEach((item) => {
-      let theItem
-      const itemName = item.id.value.slice(10)
-      if (mcData.itemsByName[itemName]) {
-        theItem = mcData.itemsByName[itemName]
-      } else {
-        theItem = mcData.blocksByName[itemName]
-      }
+      const itemName = item.id.value.slice(10) // skip game brand prefix
+      const theItem = mcData.itemsByName[itemName] || mcData.blocksByName[itemName]
```
zardoy, line 68: just refactored thats it

src/lib/plugins/login.js
```diff
@@ -234,11 +226,15 @@ module.exports.player = async function (player, serv, settings) {
...
     announceJoin()
+    // mineflayer emits spawn event on health update so it needs to be done as last step
+    player.updateHealth(player.health)
     player.emit('spawned')
 
     await player.waitPlayerLogin()
     player.sendRestMap()
     sendChunkWhenMove()
+
+    player.save()
```
zardoy, line 237: by doing this a new entry in playerdata/ appears immediately after a new player joins

src/lib/plugins/world.js
```diff
@@ -143,6 +150,10 @@ module.exports.server = async function (serv, { version, worldFolder, generation
 }
 
 module.exports.player = function (player, serv, settings) {
+  player.save = () => {
```
zardoy, line 152: refactor playerDat.save -> player.save to expose it

src/lib/plugins/world.js
```diff
@@ -243,6 +253,12 @@ module.exports.player = function (player, serv, settings) {
       .catch((err) => setTimeout(() => { throw err }), 0)
   }
 
+  // todo as I understand need to handle difficulty packet instead?
+  player.onPlayerChangeRenderDistance = (newDistance = player.view, forced = false) => {
```
rom1504, line 257: Why this function instead of listening to an event ?

src/lib/plugins/world.js
rom1504: For any added function please add to doc

src/lib/plugins/world.js
```diff
@@ -41,6 +43,10 @@ module.exports.server = async function (serv, { version, worldFolder, generation
   serv.netherworld = new World(generations.nether(generation.options))
   // serv.endworld = new World(generations["end"]({}));
 
+  // todo as i understand this is only thing that should block connection
+  serv.emit('worldsReady')
```
rom1504, line 47: Please add to doc

src/lib/plugins/login.js
rom1504: What about instead checking if it's defined where it's used ?

src/lib/plugins/log.js
```diff
@@ -1,20 +1,39 @@
...
+    input: process.stdin,
+    output: process.stdout
+  })
+
+  rl.setPrompt('> ')
+  rl.prompt(true)
+}
 
 module.exports.server = function (serv, settings) {
+  _servers.push(serv)
```
rom1504, line 35: What is _servers ?

src/lib/plugins/log.js
rom1504: I understand the intent and I agree the feature is nice. However that couple the lib and the app parts of flying-squid
What about instead define this as some ServerManager class and use it in app.js ?

src/index.js
rom1504: How and by what are options changed? Are you sure that happens before options are used ?

rom1504:
Overall looks good, but left some comments

src/lib/plugins/world.js
zardoy: To address the comment:
I think its better to add `pluginsReady` property and event so we can create other async plugins that should block connections later (for now only builtin world.js is async). However I'm not sure what to throw in case if player connects too early like `Plugins are still initializing, try again later`?

src/index.js
zardoy: Some plugin's options are used inside callbacks (`view-distance` is one of these)
Anyway we can already modify the original options object, this is just done for convenience, so I can also use options value like this:

```js
const server = createMCServer(...)
server.options.worldFolder // do smth with this
```

src/lib/plugins/log.js
zardoy: I thought it was possible to create multiple servers. However, now I understand that I could just inline exit-hook handler here.

src/lib/plugins/log.js
zardoy: What about exit-hook: this is totally okay to have it as part of the library and why not have it so "better behavior is enabled by default"?

zardoy: Hm, as I understand there is also no way to disable readline, right?

src/lib/plugins/login.js
zardoy: I removed mocking `remoteAddress` since nothing will crash if it is `undefined`, but maybe it should be documented in some way I don't really know (so other plugins can know it can be undefined)

src/lib/plugins/world.js
zardoy: Changed to event. I dont mind having it as event.

src/lib/plugins/world.js
zardoy: Done. see latest changes

src/index.js
rom1504: Are you sure that happens before options are used ?

src/lib/plugins/log.js
rom1504: "exit-hook: this is totally okay to have it as part of the library"

no that's exactly the part I don't like.

I think it's good if the library exposes a function that makes this possible, and then it's called by default in https://github.com/PrismarineJS/flying-squid/blob/master/app.js

having the library actually call it forces the user into a specific application setup

src/lib/plugins/log.js
rom1504: why is disabling date display related with being in node or not ? what about making this an option instead ?

src/lib/plugins/login.js
rom1504: where did you remove it ? I don't understand

src/lib/plugins/login.js
rom1504: what about returning instead of throwing if this is not an error ?

zardoy: Yes of course it makes sense I fixed it here https://github.com/PrismarineJS/flying-squid/pull/633/commits/923a8296343de7daf4a31d08741b58517119c656

src/index.js
zardoy: Which options do you mean? Some options such as `view-distance` are used when the player connects for example (sendMap, sendRestMap). However, I mostly used it for reading options only. I don't mind removing it if you don't like it tho.

src/lib/plugins/log.js
zardoy: > no that's exactly the part I don't like.

Alright, I don't mind making it an opt-in behavior, however, then I think it would make sense to make creating readline interface optional as well. What do you think?
Otherwise I might be missing use-cases of this library (sorry)

> I think it's good if the library exposes a function that makes this possible

Will extract, doc later today or tomorrow, not a problem

src/lib/plugins/login.js
zardoy: Yeah, sorry for now making it clear I meant this https://github.com/PrismarineJS/flying-squid/pull/633/commits/18fe14a6d295cf6da45920eb12432aa89421698f . However I removed mocking : https://github.com/PrismarineJS/flying-squid/pull/633/commits/6dff61f3868eebecbc3fabeac0623d5796aaf538 is it better?
I'm open to any option here.

src/lib/plugins/log.js
zardoy: > I think it's good if the library exposes a function that makes this possible, and then it's called by default 

Maybe this should be done as option? like `registerExitHook: true` ?

src/lib/plugins/log.js
zardoy: I added `serv.formatMessage` so I can be flexible on my side

rom1504:
overall lgtm, merged

PR #634, opened 2023-09-03
Add a few flags to cli & disable `everybody-op` by default?

zardoy:
So its easier specify IMO commonly used options for testing without having to change config


rom1504:
ci fails

PR #635, opened 2023-09-03
tab complete & `/give`, `/effect` improvements

zardoy:
/give:
1. can use name in any form
2. issues message if an item is not found so the player understands why an item is not given
3. works with non-empty inventories

- completions for other commands (items, blocks, entities...)


zardoy:
Tab complete needs fixes. It doesn't work with 1.16 at all

PR #645, opened 2024-01-04
Add 1.16.5 support and future-proof version system

Pandapip1:
~~Blocked by: https://github.com/PrismarineJS/minecraft-data/pull/823; new version of node-minecraft-data~~ Done!

Forked from #599. The author appears to have gone AWOL.

~~The goal of this PR is to add version support up to 1.18.2~~

Never mind. That will take a lot more time. For now, this has stable 1.16.5 support.

Fixes https://github.com/PrismarineJS/flying-squid/issues/447 and the following unreported minor issues:

- Duplicate interaction registration on startup
- Error when a client quits before chunks are loaded
- Error when a client quits, at all

Closes #599


src/lib/features.json
```diff
@@ -2,131 +2,136 @@
   {
     "name": "unloadChunkByEmptyChunk",
     "description": "Chunk unloading is done by sending an empty chunk",
-    "versions": ["1.8"]
+    "versions": ["1.8", "1.8.1", "1.8.2", "1.8.3", "1.8.4", "1.8.5", "1.8.6", "1.8.7", "1.8.8"]
```
rom1504, line 5: instead of doing that, what do you think about migrating this file to minecraft data support version which can handle ranges https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/common/features.json ?

src/index.js
```diff
@@ -46,10 +46,11 @@ class MCServer extends EventEmitter {
 
   connect (options) {
     const version = require('minecraft-data')(options.version).version
-    if (!supportedVersions.some(v => v.includes(version.majorVersion))) {
+    if (!supportedVersions.some(v => v.includes(version.minecraftVersion))) {
```
rom1504, line 49: I recommend we replace that by the same system as in mineflayer https://github.com/PrismarineJS/mineflayer/blob/master/lib/loader.js#L111-L120

src/index.js
```diff
@@ -46,10 +46,11 @@ class MCServer extends EventEmitter {
...
       throw new Error(`Version ${version.minecraftVersion} is not supported.`)
     }
-    this.supportFeature = feature => supportFeature(feature, version.majorVersion)
+
+    this.supportFeature = feature => supportFeature(feature, version.minecraftVersion)
```
rom1504, line 53: let's replace that by node-minecraft-data supportFeature

src/lib/version.js
```diff
@@ -1,3 +1,3 @@
 module.exports = {
-  supportedVersions: ['1.8.8', '1.9.4', '1.10.2', '1.11.2', '1.12.2', '1.13.2', '1.14.4', '1.15.2', '1.16.1']
+  supportedVersions: ['1.8', '1.8.1', '1.8.2', '1.8.3', '1.8.4', '1.8.5', '1.8.6', '1.8.7', '1.8.8', '1.9', '1.9.1', '1.9.2', '1.9.3', '1.9.4', '1.10', '1.10.1', '1.10.2', '1.11', '1.11.1', '1.11.2', '1.12', '1.12.1', '1.12.2', '1.13', '1.13.1', '1.13.2', '1.14', '1.14.1', '1.14.2', '1.14.3', '1.14.4', '1.15', '1.15.1', '1.15.2', '1.16', '1.16.1', '1.16.2', '1.16.3', '1.16.4', '1.16.5']
```
rom1504, line 2: I think it would be better here to put the testedVersions, and have that be at maximum a few per major
and then the actually supported versions is the range from the min to the max supported version, see https://github.com/PrismarineJS/mineflayer/blob/master/lib/version.js as a model

rom1504:
Thank you for working on that.

I left some comments, mostly about following the same conventions we now follow in other PrismarineJS projects and I think will help here.

src/lib/features.json
Pandapip1: Sure thing.

src/index.js
Pandapip1: That seems reasonable.

Pandapip1:
Sounds good. There's a test case that's failing. I don't know why. It works on my machine :shrug:

src/index.js
Pandapip1: The link you sent was for ensuring that the client and server are using compatible protocols, not for ensuring that the version is supported by mineflayer.

src/index.js
```diff
@@ -45,11 +43,22 @@ class MCServer extends EventEmitter {
...
 
   connect (options) {
-    const version = require('minecraft-data')(options.version).version
-    if (!supportedVersions.some(v => v.includes(version.majorVersion))) {
-      throw new Error(`Version ${version.minecraftVersion} is not supported.`)
+    const mcData = require('minecraft-data')(options.version)
+    const version = mcData.version
+    const { 0: maxMajor, 1: maxMinor, 2: maxPatch = 0 } = latestSupportedVersion.split('.')
+    const { 0: minMajor, 1: minMinor, 2: minPatch = 0 } = oldestSupportedVersion.split('.')
+    const { 0: major, 1: minor, 2: patch = 0 } = version.minecraftVersion.split('.')
```
rom1504, line 50: this will not work for snapshot version. Why not use the same logic as in mineflayer?

Pandapip1: You linked to the wrong file; I wasn't able to easily find it so I wrote my own.

I'll go look again, but harder.

Pandapip1: Never mind. The naming confused me. I'll implement it now.

src/index.js
rom1504:
```
    const versionData = bot.registry.version
    if (versionData['>'](latestSupportedVersion)) {
      throw new Error(`Server version '${serverPingVersion}' is not supported. Latest supported version is '${latestSupportedVersion}'.`)
    } else if (versionData['<'](oldestSupportedVersion)) {
      throw new Error(`Server version '${serverPingVersion}' is not supported. Oldest supported version is '${oldestSupportedVersion}'.`)
    }
```
this is what checks supported versions in mineflayer

src/lib/plugins/placeBlock.js
```diff
@@ -60,23 +65,6 @@ module.exports.server = (serv, { version }) => {
       }
       return data
     }
-
-    // Register default handlers for item -> block conversion
-    for (const name of Object.keys(mcData.itemsByName)) {
-      const block = mcData.blocksByName[name]
-      if (block) {
-        if (block.states.length > 0) {
-          serv.onItemPlace(name, ({ properties }) => {
```
rom1504, line 69: did you check this is used nowhere?

src/lib/plugins/respawn.js
```diff
@@ -12,7 +12,7 @@ module.exports.player = function (player, serv) {
       player.behavior('requestRespawn', {}, () => {
         player._client.write('respawn', {
           previousGameMode: player.prevGameMode,
-          dimension: serv.supportFeature('dimensionIsAString') ? serv.dimensionNames[0] : 0,
+          dimension: (serv.supportFeature('dimensionIsAString') || serv.supportFeature('dimensionIaAWorld')) ? serv.dimensionNames[0] : 0,
```
rom1504, line 15: dimensionIaAWorld typo

src/lib/plugins/world.js
```diff
@@ -290,7 +291,7 @@ module.exports.player = function (player, serv, settings) {
     }
     player._client.write('respawn', {
       previousGameMode: player.prevGameMode,
-      dimension: serv.supportFeature('dimensionIsAString') ? serv.dimensionNames[opt.dimension || 0] : opt.dimension || 0,
+      dimension: (serv.supportFeature('dimensionIsAString') || serv.supportFeature('dimensionIaAWorld')) ? serv.dimensionNames[opt.dimension || 0] : opt.dimension || 0,
```
rom1504, line 294: dimensionIaAWorld typo

src/index.js
```diff
@@ -45,11 +43,19 @@ class MCServer extends EventEmitter {
   }
 
   connect (options) {
-    const version = require('minecraft-data')(options.version).version
-    if (!supportedVersions.some(v => v.includes(version.majorVersion))) {
-      throw new Error(`Version ${version.minecraftVersion} is not supported.`)
+    const mcData = require('minecraft-data')(options.version)
+    const serverPingVersion = mcData.version
```
rom1504, line 47: this variable name should be different, for example version

src/index.js
```diff
@@ -45,11 +43,19 @@ class MCServer extends EventEmitter {
   }
 
   connect (options) {
-    const version = require('minecraft-data')(options.version).version
-    if (!supportedVersions.some(v => v.includes(version.majorVersion))) {
-      throw new Error(`Version ${version.minecraftVersion} is not supported.`)
+    const mcData = require('minecraft-data')(options.version)
```
rom1504, line 46: no need to use both prismarine-registry and mcdata

prismarine-retry is a replacement of mcdata

you can use only prismarine-registry

src/index.js
```diff
@@ -45,11 +43,19 @@ class MCServer extends EventEmitter {
...
+    const serverPingVersion = mcData.version
+    const registry = require('prismarine-registry')(serverPingVersion)
```
rom1504, line 48: this should take as input options.version

src/index.js
```diff
@@ -45,11 +43,19 @@ class MCServer extends EventEmitter {
...
+
+    const versionData = registry.version
+    if (versionData['>'](latestSupportedVersion)) {
+      throw new Error(`Server version '${serverPingVersion}' is not supported. Latest supported version is '${latestSupportedVersion}'.`)
+    } else if (versionData['<'](oldestSupportedVersion)) {
+      throw new Error(`Server version '${serverPingVersion}' is not supported. Oldest supported version is '${oldestSupportedVersion}'.`)
     }
-    this.supportFeature = feature => supportFeature(feature, version.majorVersion)
+
+    this.supportFeature = feature => mcData.supportFeature(feature)
```
rom1504, line 58: this can use registry.supportFeature

rom1504:
left some more comments but most things look good besides that

rom1504:
would be good if we can finally make progress on version supports for flying squid

src/lib/plugins/placeBlock.js
Pandapip1: Yes. It was used to generate part of the stuff that's now in minecraft-data.

Pandapip1:
We've got a green!

src/lib/plugins/placeBlock.js
rom1504: https://github.com/search?q=repo%3APrismarineJS%2Fflying-squid+serv.onItemPlace%28&type=code it seems used in many places

rom1504:
Left one comment about that onItemPlace

src/lib/plugins/settings.js
```diff
@@ -7,11 +7,11 @@ function randomInt (low, high) {
 module.exports.server = function (serv, settings) {
   serv.gameMode = settings.gameMode
   serv.difficulty = settings.difficulty
-  const mcData = require('minecraft-data')(settings.version)
+  const registry = require('prismarine-registry')(settings.version)
```
extremeheat, line 10: There should really be one instance of prismarine-registry, probably under server.registry, as the point of prismarine-registry is to create a copy of Minecraft data that can be added onto (with login codec data or other things). Otherwise calling methods on the registry will only update the data for that specific registry instance

src/lib/plugins/placeBlock.js
Pandapip1: Oh whoops I completely misread the file name. I thought this was the file I deleted.

The `serv.onItemPlace` method is not deleted and there are no side effects to moving the default handlers to the function as far as I am aware.

https://github.com/PrismarineJS/flying-squid/blob/d11b23d8a606c86202127ccd60bd3f023339a8ed/src/lib/plugins/placeBlock.js#L17-L25

src/lib/plugins/settings.js
Pandapip1: I've changed enough things already and declare that out of scope for this PR. 

Also, the registries might be using different versions because player plugins create the registry using the player protocol version.

rom1504:
Thanks for the PR.

There's a pretty big pile of new versions to get this to current: 1.17, 1.18, 1.19, 1.20
You can look at corresponding mineflayer PR to know what needs to be changed / implemented

PR #654, opened 2024-03-29
fixed gramma

privt00:
this is now the right one i forgot to change the name in the old one


privt00:
how can this check fail its a word that got changed

PR #656, opened 2024-05-07
Update portal checking to ensure clicking on another block next to the portal activate it.

emotikon90elo5:
OLD: old
NEW: new


rom1504:
please instead fix the detect frame function

rom1504:
Old and not ready

PR #660, opened 2024-12-31
Fix spawning in 1.16.5 worlds, 1.17

extremeheat:
* Fix missing spawn packets causing player to become glitched/world not render
* Fixes to registry use

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -7,14 +7,10 @@ if (typeof process !== 'undefined' && !process.browser && process.platform !== '
 
 const { createServer } = require('minecraft-protocol')
 
-const EventEmitter = require('events').EventEmitter
+const { EventEmitter } = require('events')
 const { testedVersions, latestSupportedVersion, oldestSupportedVersion } = require('./lib/version')
 const Command = require('./lib/command')
 const plugins = require('./lib/plugins')
-require('emit-then').register()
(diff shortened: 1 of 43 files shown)


src/index.js
```diff
@@ -23,7 +19,6 @@ module.exports = {
   generations: require('./lib/generations'),
   experience: require('./lib/experience'),
   UserError: require('./lib/user_error'),
-  portal_detector: require('./lib/portal_detector'),
```
rom1504, line 26: Why is this removed?

rom1504: Got it

rom1504:
I think it could be nicer to first do a PR changing the registry stuff with no change on behavior then another PR doing the 1.17 and 1.16.5 fixes

But I guess I'm also ok to merge like this

So more important question: does it work ?
For example 
* Loading/saving the world
* Basic features like dig build, move around,  commands

extremeheat:
Commands were unchanged 1.16-1.18

Movement, dig, break and world saving are working 1.16 - 1.18 (the mask issue should only have been a problem when sending over network)

I'll open another PR after this one for 1.18, not many changes needed there. For 1.16.5 the main issue seems to have been missing status packets

rom1504:
Ok

Issue #649, opened 2024-01-08
Fix failing block-related tests


Pandapip1:
The 'place a block', 'can open and close a chest', and/or `/setblock` tests inconsistently fail for no apparent reason.

rom1504:
https://github.com/PrismarineJS/flying-squid/actions/runs/12569838389/job/35038952697

rom1504:
looks fixed

extremeheat:
Seems the tests are mostly failing on the "before each" hook at not the tests themselves, and --retries doesn't work on beforeEach

rom1504:
yeah I guess we should fix the root cause...

PR #662, opened 2025-01-02
1.18 and 1.18.2 support

extremeheat:
Try to address #649 

After https://github.com/PrismarineJS/prismarine-chunk/pull/267

### DIFF
--- a/src/lib/plugins/world.js
+++ b/src/lib/plugins/world.js
@@ -4,6 +4,7 @@ const generations = require('../generations')
 const playerDat = require('../playerDat')
 const spiralloop = require('spiralloop')
 const { level } = require('prismarine-provider-anvil')
+const nbt = require('prismarine-nbt')
 
 module.exports.server = async function (serv, options = {}) {
   const { version, worldFolder, generation = { name: 'diamond_square', options: { worldHeight: 80 } } } = options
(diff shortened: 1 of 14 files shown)


src/lib/plugins/world.js
```diff
@@ -237,23 +246,17 @@ module.exports.player = function (player, serv, settings) {
...
       .map(t => ({
         chunkX: playerChunkX + t[0] - view,
         chunkZ: playerChunkZ + t[1] - view
       }))
       .filter(({ chunkX, chunkZ }) => serv._worldLoadPlayerChunk(chunkX, chunkZ, player))
-      .reduce((acc, { chunkX, chunkZ }) => {
-        const p = acc
-          .then(() => player.world.getColumn(chunkX, chunkZ))
-          .then((column) => player.sendChunk(chunkX, chunkZ, column))
-        return group ? p.then(() => sleep(5)) : p
```
rom1504, line 250: Removing this logic will cause nmp server to kill the vanilla client by sending chunks to fast.

extremeheat: Ah i see, reverted. It seemed to not be waiting for all the promises to resolve before spawning the player into the world causing it to sometimes fall through the world but that may be something else

rom1504:
is this ready to merge?

rom1504:
the code lgtm

rom1504 (approved):

PR #663, opened 2025-01-04
1.19

extremeheat:
* Support 1.19, 1.19.2, 1.19.3, 1.19.4
* Disable sending tablist packet as that depends on key signing and that the structure is too messy between versions (this will be fix in https://github.com/PrismarineJS/minecraft-data/pull/948 but will need combined changes to flying-squid and mineflayer)
* Chat signing not currently implemented, all messages from server to client are sent as system_chat messages which cannot be reported.  Not important for gameplay but can be added for correctness/debugging/tab packet later
* After https://github.com/PrismarineJS/minecraft-data/pull/956


rom1504:
based on https://github.com/extremeheat/flying-squid/pull/1 (see for details)

rom1504:
opened https://github.com/PrismarineJS/flying-squid/issues/665 and https://github.com/PrismarineJS/flying-squid/issues/664 for follow up on chat signing and tablist

src/lib/plugins/login.js
```diff
@@ -175,13 +200,10 @@ module.exports.player = async function (player, serv, settings) {
         ping: otherPlayer._client.latency
       }))
     })
-    setInterval(() => player._client.write('player_info', {
```
rom1504, line 178: looks like this is completely removed even for before 1.19 ; is this expected?

extremeheat: This sends the latency for all the clients but the previous player_info packet already sends the latencies in its payload with action mode 0

Back to 1.8 https://github.com/PrismarineJS/minecraft-data/blob/master/data/pc/1.8/proto.yml#L565

I think it may have been added because it's what vanilla did at some point but it wasn't doing that for 1.17. I could re-add it behind a flag for <1.17

extremeheat: Done

PR #669, opened 2025-01-05
1.20.3, 1.20.5 and 1.21.1 support

extremeheat:
After https://github.com/PrismarineJS/node-minecraft-protocol/pull/1364


extremeheat:
flying-squid updated to 1.21.1. The remaining flaky tests are pretty much all due to zlib. I tried debugging this a bit but it's hard as it's not 100% clear exactly where the zlib errors are happening. As the zlib errors are thrown in c++ there is no JS stack context to rely on which makes it a pain, and also that they're async means it's not easy to log before/around all the zlib calls as something could be going on with the event loop.

I think we could maybe move all the zlib to sync for some extra performance in the future. But for now it's not a big deal

rom1504:
Nice!

PR #688, opened 2025-09-07
Updated the app.js code and README.md files so that PrismarineJS can be used under the latest NodeJS versions

AyzeLYC:
Under the latest NodeJS versions if you want to use require you have to use :

import { createRequire } from "node:module"; // imports the createRequire function from the NodeJS CommonJS module handler
const require = createRequire(import.meta.url); // initializes the require function that NodeJS had under CommonJS

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,3 +1,7 @@
+// global dependancies
+import {createRequire} from "node:module";
+const require = createRequire(import.meta.url);
+
 #!/usr/bin/env node
 // process.env.DEBUG = 'minecraft-protocol'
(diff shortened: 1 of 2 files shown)


extremeheat:
This is not correct

extremeheat:
Update your node to latest if you have issue importing the package with ESM

extremeheat:
flying-squid (as with all the other prismarinejs projects) are commonjs, not ESM. You cannot use ESM with them. If you are on a newer version of node and using ESM, you can import commonjs with a standard ESM import.

